PR #5, opened 2017-02-11
Updated Package.appxmanifest sample

qmatteoq:
Updated the Package.appxmanifest sample, since without the PhoneIdentity element it isn't possible to build the WinJS project and use it to generate the app packages using the Store -> Create App Packages option in Visual Studio


awkoren:
Adding @rido-min for review.

rido-min (approved):
Looks good to me

awkoren (approved):

PR #7, opened 2017-02-15
Proposed removal of unformatted "Important APIs"

riverar:
There's a weak link of what's in this list to the page its on, seems cleaner to just remove it (especially given it's not styled correctly).

Screenshot:
image


muhsinking:
Hi @riverar thanks for pointing this out--it looks like the issue persists across all but one of the "Enumerate devices" topics. For now I'm going to keep the link and fix the formatting.

PR #16, opened 2017-03-02
Update create-a-basic-windows-10-app-in-cpp.md

tylerwowen:
`\"` doesn't work. `&quot;` is the way to add double quote.


GrantMeStrength:
Thanks, I've fixed this now - the PR isn't taken as so much else changed on the page recently I didn't want to break anything. Your input is much appreciated!

tylerwowen:
`\"` appear several different places in the page. You may want to update those as well.

PR #21, opened 2017-03-14
Fix typo on line 127

TheSpecialist4:
Fixed the typo on line 127 from "opent he ..." to "open the ..."


mikematteson (approved):
Thanks for spotting and fixing this. The change should be visible online in a couple of days.

PR #24, opened 2017-03-18
Fix bug in samples using isCtrlKeyPressed field

zholobov:
Using isCtrlKeyPressed field to track Ctrl key down state is a bug. For example, if user holds Ctrl key and then switches focus to another window (by clicking mouse outside of current window, or by other means, or if focus changes without user consent when another app steals focus) the current window will not get KeyUp. And when user subsequently releases Ctrl and then switches focus back to the window - the isCtrlKeyPressed field will remain true (until user presses and releases Ctrl again in this window) so if user then presses P in this code sample without actually holding Ctrl the code will interpret it as Ctrl+P and will carry out the command.

A proper solution is to check for Ctrl being pressed on each KeyUp right before making decision on the command.

### DIFF
--- a/windows-apps-src/input-and-devices/keyboard-interactions.md
+++ b/windows-apps-src/input-and-devices/keyboard-interactions.md
@@ -432,7 +432,7 @@ The following example demonstrates how to implement shortcut keys. In this examp
 Note also that the page sets input focus to itself when it is loaded. Without this step, no control has initial input focus, and the app does not raise input events until the user sets the input focus manually (for example, by tabbing to or clicking a control).
 
 ```xaml
-<Grid KeyDown="Grid_KeyDown">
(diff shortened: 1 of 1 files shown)


zholobov:
C++ and VB.NET parts of sample may require proper syntax, sorry for that.

msftgits:
Microsoft employee validated, no CLA required

Karl-Bridge-Microsoft (commented):
Changes look good. However, topic is being updated significantly, and this code has been moved into another topic.

Karl-Bridge-Microsoft:
Topic has been revised significantly and is waiting to be pushed to live. I've pointed the feature team to this PR to look at and assess your suggested updates. If approved, we'll migrate those changes to the new topic. Thanks for your contributions!

PR #28, opened 2017-03-24
Update README.md

yinyue200:
"##Title" is not a correct markdown code in Github now


mikematteson (approved):
You're right. Thanks for the help!

PR #30, opened 2017-03-25
fixed html escape codes in code samples

ThomasKramer:
Some code samples were broken due to unnecessary  html escape codes.
I replaced the escape codes with the proper characters.


mikematteson (approved):
Thanks for spotting these issues and offering fixes for them, Thomas. The individual topic owners have reviewed and in most cases accepted your changes. Much appreciated!

PR #32, opened 2017-03-28
Fix spacing in how-to-create-and-consume-an-app-service.md

cfeilen:
Fixes an extra space which was added to lines 194-256.  This extra space caused formatting issues in the display of the .MD file.


TylerMSFT (approved):
Thank you.

PR #33, opened 2017-03-28
Remove Lock Screen Apps Guidance

cc-cortes:
Lock Screen Apps and use of Background Tasks for them were an 8.1 concept. They are not a concept in Windows 10 and the section around guidelines for them should be removed from the Background Task documentation.


TylerMSFT (approved):
This section doesn't say that you have to

TylerMSFT:
Thanks Chris. Someday we may stop finding references to this lock screen stuff.

PR #36, opened 2017-03-30
Add warning about ExtendedExecutionForegroundSession

cc-cortes:
The ExtendedExecutionForegroundSession and ExtendedExecutionForegroundReason are not available for public consumption. Need to update the documentation to let developers be aware of that.


TylerMSFT (approved):
Thanks, Chris.  incorporated.

PR #40, opened 2017-04-03
Initialize _displayRequest

darenm:
The _displayRequest variable is never initialized.


drewbatgit:
Thanks for catching this. I already have a fix for this in a private branch that will be merged into the live branch this week. Instead of checking for null before setting the value, I initialize the object using a field initializer when the DisplayRequest object is declared.

PR #46, opened 2017-04-05
Identify Suspending as the best place to save data

cc-cortes:
Change the wording so that suspending is still the best place to save data. Add the caveat that if you are doing background work, it is also good to perform a save operation in EnteredBackground.


TylerMSFT (approved):
Thanks, Chris.

PR #50, opened 2017-04-06
POS Documentation Typos

theREALjdeck:
Espon needs to be altered to Epson
darcode needs to be altered to barcode


mikematteson (approved):
Thanks for your help in spotting and fixing these. Much appreciated!

PR #53, opened 2017-04-08
Added a notice about relative file paths

sylveon:


laurenhughes (approved):
Thank you for taking the time to add clarification about MakeAppx!

laurenhughes (approved):
Thank you for taking the time to add clarification about MakeAppx!

PR #57, opened 2017-04-12
Update run-a-background-task-on-a-timer-.md

TylerMSFT:
The formatting is broken. Pulling the line back to see if the code block will then render properly.
See https://docs.microsoft.com/en-us/windows/uwp/launch-resume/run-a-background-task-on-a-timer-


TylerMSFT:
Leave a comment

TylerMSFT:
Trying this a different way.

PR #59, opened 2017-04-14
Update gatt-server.md

chandrakantnitt:
Fixed the title as it was showing up as GATT Client - which ideally should be GATT Server


chandrakantnitt:
@mijacobs  Could you please review this ?

PR #66, opened 2017-04-23
Remove a duplicate sentence

mcooley:
Also, remove leading ">" characters in two places, since this style isn't used elsewhere in the article.


TylerMSFT (approved):
Thanks, Matt

PR #67, opened 2017-04-24
Update choose-a-uwp-version.md

kraigb:
The intro text said "three versions" whereas the table has four. I edited the line to introduce the table instead and eliminate a specific number.


QuinnRadich:
Kraig - thanks for the contribution. This change happens to align with other work to the page that I'm doing today. I'm going to close this without merging, but the fix will still go live tomorrow.

Thank you!

PR #69, opened 2017-04-24
Delete unecessary ending tags

rootasjey:


mcleblanc (approved):
Thanks for spotting these extraneous tags and making the corrections. Your contribution is much appreciated!

PR #88, opened 2017-05-12
Fix discrepancy in code explanation

faheel:
`Height` of a `RowDefinition` is `2*` in the code snippet, which is also how the image shows it, whereas the explanation given under it mistakenly says that it is `3*`.


muhsinking (approved):
Hey Faheel, thanks so much for catching this!

PR #97, opened 2017-05-16
remove EnteredBackground, add EE to Suspend page

cc-cortes:
Remove advice about using EnteredBackground as the location for saving user state, it is confusing developers. It has already been removed from the App Lifecycle page. Also add information about using Extended Execution to extend the Suspending state.


TylerMSFT (approved):
Thanks Chris.

PR #108, opened 2017-05-24
Update sensor-orientation.md

JocelynBerrendonner:
Fixing a confusing sentence. The sensor axis are not always defined based on landscape orientation of the device. As per explained in our detailed documentation: https://msdn.microsoft.com/en-us/library/windows/hardware/dn642102(v=vs.85).aspx, axis are always related to the device natural orientation. Y will always point up and X to the right when a device is held upstraight in its natural orientation (which typically is landscape for a laptop and portrait for a phone)


muhsinking (approved):
Hey Jocelyn, thanks for catching this! Hope you have a great day 👍

PR #116, opened 2017-06-01
Add Remote Deployment to PC

hpsin:
cc @cdonthini and @PatrickFarley

### DIFF
--- a/windows-apps-src/debug-test-perf/deploying-and-debugging-uwp-apps.md
+++ b/windows-apps-src/debug-test-perf/deploying-and-debugging-uwp-apps.md
@@ -68,7 +68,7 @@ To return to this dialog, you can open project properties and go to the **Debug*
 
 ![Debug tab](images/debug-remote-machine-config.png)
 
-To deploy an app to a remote PC, you will also need to download and install the Visual Studio Remote Tools on the target PC. For full instructions, see [Remote PC instructions](#remote-pc-instructions).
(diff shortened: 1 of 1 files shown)


windows-apps-src/debug-test-perf/deploying-and-debugging-uwp-apps.md
```diff
@@ -172,10 +175,11 @@ When you first register the layout from the network, your credentials will be ca
 You cannot select **keep all files on device** when you register the layout from the network because no files are physically copied to the remote device.
 
 > [!NOTE]
-> **Register layout from network** is currently supported on Xbox running Windows 10 Anniversary Update.
+> **Register layout from network** is currently supported on Xbox running Windows 10 Anniversary Update and PCs running Windows 10 Creators Update.
 
 On the remote device, the layout gets registered to the following default location depending on the device family:
-  `Xbox: \\MY-DEVKIT\DevelopmentFiles\XrfsFiles`
+  `Xbox: \\MY-DEVKIT\DevelopmentFiles\XrfsFiles` - this is a symlink to the **package registration path**
```
c-don, line 181: do we want to say stuff about the symlink. Maybe there is a reason, Xbox avoided that. We can just say- On PC, we register to the package registration path. what do you think

c-don (changes requested):
looks good! One small comment regarding the symlink.

windows-apps-src/debug-test-perf/deploying-and-debugging-uwp-apps.md
hpsin: Fine as is, it's a symlink, no need to hide that. Clarifies why it looks like files are being copied to disk.

PatrickFarley (approved):

PR #117, opened 2017-06-02
Enable contributor display for VSTS

mattwojo:
Based on Ted Hudek email re: resolution with VS China on representing profiles for contributions made through VS Git


mattwojo:
Moving this to VSTS

PR #122, opened 2017-06-13
Add header for optimizing bg energy usage

cc-cortes:
Point to the Optimize background activity page


PatrickFarley:
The optimize-background-activity doc doesn't live in the launch-resume section, so we typically wouldn't place it in the index there. I will ask if an exception should be made, since it pertains to both categories.

PatrickFarley (approved):

PR #124, opened 2017-06-17
Update DimissExtendedSplash method

colinkiama:
Currently, an error will occur in that method saying that the code is running on a thread it's not supposed to run on. This error occurs when calling the method inside DismissedEventHandler. Using the core dispatcher eliminates the issue.


TylerMSFT:
Hi, I'm a writer at Microsoft and I appreciate your input on this topic. I ran through the steps and everything works for me without marshalling the call to dismiss the extended splash screen. I'd like to get some more info. Are you doing something slightly different than outlined in the steps? What version of the OS are you using?

TylerMSFT:
I've decided to take this.

TylerMSFT (approved):

PR #131, opened 2017-06-29
Removed 1 space so asterisk work properly

TerryWarwick:


mikematteson (approved):
Thanks for the fix. Much appreciated!

PR #150, opened 2017-07-20
Updated info on Target Version requirements

ptorr-msft:
Made it clear that the `SupportsMultipleInstances` attribute was not available on 14393 but that you had to target 15063 (which in turn requires VS 2017) if you wanted to use it.


TylerMSFT (changes requested):
Can you remove the reference to '2015 or 2017'  Let's just say 'Visual Studio' because soon it will be out of date when the next version is introduced.

TylerMSFT (approved):

PR #156, opened 2017-07-25
Transcoding does not support encode options

13thsymphony:
CreateForTranscodingAsync does not have an overload that supports encode options. I don't know how to find the proper MSDN link to the method's reference.


laurenhughes (approved):
Everything looks good. Thanks!

PR #158, opened 2017-07-26
MinorMods

haripulapaka:
made some changes to the verbiage and added more sentences around x86 and ARM packages.


haripulapaka:
@laurenhughes - minor mods, can you take a look?

laurenhughes (approved):
Thanks for the additions, Hari!

PR #159, opened 2017-07-26
Maps protocol launch updates

mcooley:
Clarify recommendations for launching the Windows Maps app.

* When you specify a map view (using cp, lvl, etc.) in conjunction with a search (q, where), the map view is used as a starting point for the search but the final map view is always the best one for the search results. Remove examples which suggest that you can control the zoom level of the final map view.
* Remove commentary about the predictability of using cp versus bb; for modern versions of the Maps app the predictability of both methods is similar.
* Rephrase the "display known locations" section to better convey that rad, hdg, and pit can be used for all styles, not just 3D views.
* Deemphasize the difference between aerial and 3D views in introductory text, since they do the same thing on Windows 10.
* Minor phrasing updates.


TylerMSFT (approved):
Thanks, Matt!

PR #161, opened 2017-07-28
Update app-package-requirements.md

diverdan92:
Adding some clarifying information about app association with the store.


JnHs (approved):
Looks good.

PR #169, opened 2017-08-04
Replace code snippets headers

AzatTazayan:
Replaced "ManagedCPlusPlus" to "cpp" , "CSharp" to "cs" and "VisualBasic" to "vb"


Karl-Bridge-Microsoft (approved):
Thanks for your contribution! Much appreciated!

PR #175, opened 2017-08-10
Add lock screen ability to Location Tracking EE

cc-cortes:
Location Tracking EE is capable of running indefinitely under the lock screen. This should be identified in the documentation.


TylerMSFT (approved):
Thanks, Chris! Appreciate your care for the docs.

PR #178, opened 2017-08-12
Update my-people-sharing.md

SergeyKn:
Fixed another typo and args.ShareOperation.Contacts is an object of IReadOnlyList type which has no Size property. Replaced with Count property and tested


muhsinking (approved):
Thanks for fixing these issues Sergey!

PR #185, opened 2017-08-18
ASLR test listed as only being performed on apps written in managed code

Wack0:
...which makes no sense. This PR changes the text to say that it's performed only on apps written in unmanaged languages, which seems to be what was initially meant, given that the /DYNAMICBASE linker option is referred to.


PatrickFarley:
I'll loop in a PM to confirm that this new meaning is correct.

PatrickFarley (approved):

PR #186, opened 2017-08-18
Fix Windows version number

peterbozso:


Karl-Bridge-Microsoft (approved):
Thanks for your contribution to the docs! It's greatly appreciated.

PR #190, opened 2017-08-28
Update universal-application-platform-guide.md

devfables:
Updated grammar and content structure


TylerMSFT (changes requested):
Hi,
Thank you for your interest.
Regarding: "From phones, tablets, desktop computers, Surface Hubs, Xbox consoles, to HoloLens; any or all may be chosen as a target platform"
I don't think the last part, "any or all may be chosen as a target  platform", conveys the right idea. That suggests that after you select universal that you select the devices you want to run on. Universal simply means that it will run on all devices that support the UWP platform.
Other than that, the edits are ok.

devfables:
Added an additional commit (c60dcb3) that I think should clear up the language.

TylerMSFT (approved):

TylerMSFT:
Thanks, @devfables .

PR #191, opened 2017-08-28
Add notes on disabling SSH and known SSH issue

hpsin:
@PatrickFarley

### DIFF
--- a/windows-apps-src/get-started/enable-your-device-for-development.md
+++ b/windows-apps-src/get-started/enable-your-device-for-development.md
@@ -117,12 +117,28 @@ If you encounter problems enabling Developer Mode or Device Portal, see the [Kno
 
 ###SSH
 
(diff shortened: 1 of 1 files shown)


mikematteson:
@laurenhughes @PatrickFarley Who should this PR be assigned to? Thanks!

QuinnRadich (commented):
This seems like elaboration that would be more suited for a page that covers SSH in more detail, rather than the Get Started page for enabling a device for deployment. Is there a different page in the docs we could link to instead of putting this information here?

hpsin:
Woops, sorry @QuinnRadich  - missed this.  There is no page for SSH currently, although we could create one, especially as my latest PR includes even more information about SSH.

QuinnRadich:
Understood.

We should look towards the creation of such a page if necessary, then. This Get Started page is already on the border of being a little too in-depth for most of its audience - I'd like to keep that as limited as possible.

For those reasons, I think it's not appropriate to put that information on this page at the time being. Something to link to would be much better.

PR #195, opened 2017-09-01
Delete wdp-updatepolicy-api.md

WilliamsJason:
We have better alternatives to this API, it's still marked as "coming in the next preview" so we should just deprecate it.


mikematteson:
Issues found on PR checklist review; @mtoepke Please route to appropriate CX content owner to ensure compliance with these requirements:
- Any deleted topics must be redirected to appropriate new targets.
- Broken links must not be introduced. If a topic is being deleted, all crosslinks to that topic must be deleted.

Thanks!

WilliamsJason:
Is there a way to find all crosslinks to a topic? Searching for "wdp-updatepolicy-api" didn't work. We should delete any links to this topic along with deleting the topic itself.

mtoepke:
Jason already removed the one reference to this.  It was never in the TOC and was pre-release, so it shouldn't be redirected.

PR #198, opened 2017-09-05
Small Typo Fix

kurodansei:
This article encourates, when I think it should be a bit more encouraging.


mikematteson (approved):
Thanks for your spotting this and taking time to submit a correction. We appreciate your contribution!

PR #201, opened 2017-09-14
Fix top level links in the CompositionBrush Overview

HarneetSidhana:
Updating links to be all lower case since upper case in the links don't work


jwmsft:
The links are fixed by another commit, so this PR is not needed.

PR #208, opened 2017-10-08
Added Applies To section

andreszs:
Along this entire series of articles, the person who wrote them failed to mention that 99% of these so-called "windows universal" features are **valid only for Windows 10, not Windows Phone 8.1**. The W10 universe is definitely separated from the WP8.1 one, like the Marvel multiverses.


mikematteson:
@mijacobs PR passed review checklist. Please evaluate the proposed changes further. Thanks!

mijacobs:
Thanks for the feedback, andreszs. All of our UWP docs on docs.microsoft.com are Windows 10-only. We archived our Windows 8 docs. Rather than add a note to one article, we need to find a way to make that clear in all of our articles, so I'm declining this PR for now in the interest of creating a solution that works for all of our docs.

PR #219, opened 2017-10-17
Fix bug of using a private isCtrlKeyPressed

zholobov:
Using isCtrlKeyPressed and KeyUp/Down events to track CTRL key state is wrong since it does nto cover case when user presses CTRL and without releasing it switches to another app, then releases CTRL and then returns back to our app. In this case with isCtrlKeyPressed will wrongly contain true and pressing A, P, S without any modifiers will trigger actions. Also, KeyUp is not wired in the example so CTRL will be forever stuck once pressed. The correct solution is to determine modifier key being pressed when a target key (A, S, P) is pressed by using CoreWindow.GetForCurrentThread().GetKeyState.

C# code is correct in my change. However C++ and especially VB.NET parts may be not compilable - please check and correct if needed.

### DIFF
--- a/windows-apps-src/input-and-devices/keyboard-events.md
+++ b/windows-apps-src/input-and-devices/keyboard-events.md
@@ -176,6 +176,12 @@ void KeyboardSupport::MainPage::MediaButton_Click(Platform::Object^ sender, Wind
 }
 
 
+bool KeyboardSupport::MainPage::IsCtrlKeyPressed()
+{
+    var ctrlState = CoreWindow::GetForCurrentThread().GetKeyState(VirtualKey::Control);
+    return (ctrlState & CoreVirtualKeyStates::Down) == CoreVirtualKeyStates::Down;
+}
+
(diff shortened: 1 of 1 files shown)


mikematteson:
@zholobov : Thanks for your contribution. Going forward, CX asks that Microsoft employees use VSTS to submit documentation changes. Future contributions via GitHub will be closed with a request that you resubmit them via VSTS. More information is at http://aka.ms/windowsauthoring (see **UWP instructions** > **Conceptual content**).

Cc: @Karl-Bridge-Microsoft

mikematteson:
@Karl-Bridge-Microsoft : **Passed PR checklist** Please review proposed changes further and merge if appropriate.

zholobov:
@mikematteson thank you, will use VSTS next time.

zholobov:
@Karl-Bridge-Microsoft  If you have doubts about the change I can provide a video of the bug in current sample.

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
Thanks for your contribution!

PR #228, opened 2017-10-26
Fix confusing typo

oldnewthing:
It makes no sense to call a desktop-only API when running on a mobile device. Should be "... running on a desktop device."


TylerMSFT (changes requested):
Hi Raymond,
For Microsoft folks, we make the changes in VSTS. 
I've made the change there and it will percolate to live tomorrow.
Feel free to close this request.

PR #266, opened 2017-12-06
Update typography.md

roxken:
Tiny change - removed apostrophe in "Do's" (pet peeve). Apostrophe's don't make plural's :-)


mikematteson:
@mijacobs : PR **passed** review checklist, but proposed changes may need some adjustment. Please evaluate further. Thanks!

mijacobs (commented):
This change removes a line break between the Type ramp section header and its content. Regarding do's and don'ts, style guides are mixed as to which is right, do's or dos. The rest of our content uses "do's", so we'll stick with that to be consistent.

mijacobs:
https://www.visualthesaurus.com/cm/wc/dos-and-donts-or-dos-and-donts/

PR #286, opened 2018-01-15
Default Value of Binding in UWP is always OneWay

thomasclaudiushuber:
Compared to WPF, UWP doesn't have the FrameworkPropertyMetadata.BindsTwoWayByDefault property. WPF uses that to set for example a Binding on a TextBox.TextProperty by default to TwoWay. In UWP, the FrameworkPropertyMetadata.BindsTwoWayByDefault property does not exist and there's no way to define another default mode than OneWay for a Binding. That's why the default is always one way.


pedrolamas (approved):
This is my understanding also. From my perspective, this change makes sense given the current UWP SDK.

mikematteson:
@jwmsft : PR **passed** checklist review. Please evaluate further and approve/merge if appropriate. Thanks!

jwmsft (approved):

PR #291, opened 2018-01-19
Update app-lifecycle.md

edetoc:
I believe there's an error. this sentence "Apps enter the running in background state when they are launched or activated" should be replaced with "Apps enter the running in foreground state when they are launched or activated"


mikematteson:
@TylerMSFT : PR **passed** checklist review. Please evaluate further and appove/merge if appropriate. Thanks!

TylerMSFT:
Hi edetoc.  It's actually correct as written. The app is originally launched or activated as running in the background. 
The app then gets the LeavingBackground event if the app then needs to move into the foreground due to a foreground app launch. I'll add something to clarify that.

PR #294, opened 2018-01-24
Fixing typo in sample link

strican:
The sample link was incorrectly typed "XAML Content menu sample" when it links to the "Context Menu" sample.


mikematteson (approved):
@strican : Thanks for catching this and taking time to fix it. Much appreciated!

Cc: @mijacobs

PR #299, opened 2018-02-06
fix typo

stevenbrix:
this should've been OneWay instead of OneTime. The default is OneTime *because* it generates less code and OneWay causes more to be generated


mikematteson:
@jwmsft : PR **passed** checklist review. Please evaluate the proposed change and approve/merge if appropriate.

jwmsft (approved):

PR #301, opened 2018-02-08
Update navigationview.md

JerryNixon:
PLEASE NOTE that I do not have permissions in VSTS to submit changes. That's why I am creating a pull request here. I realize it is a little inconvenient to your new process; it's all I can do.

There is a bug in the AppTitle code that does not handle FullScreenMode. I have updated the sample code to account for this. It might even be a bug that the LayoutMetricsChanged does not raise when FullScreenMode changes, and TitleBar still reserves space for the BackButton even though it is no longer visible. In either case, the code sample accounts for both scenarios.

### DIFF
--- a/windows-apps-src/design/controls-and-patterns/navigationview.md
+++ b/windows-apps-src/design/controls-and-patterns/navigationview.md
@@ -358,39 +358,45 @@ To do so, add the following code to your App.xaml.cs.
 
 ```csharp
 //draw into the title bar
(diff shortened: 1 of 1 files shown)


mikematteson:
@serenaz :  PR **passes** checklist review. Please evaluate the submission further and approve/merge if appropriate. Thanks!

serenaz (approved):
Thanks!

PR #305, opened 2018-02-14
Typo in set-app-pricing-and-availability.md

agriffard:
Remove duplicated 'the'.


mikematteson (approved):
@agriffard : Thanks for the fix!

Cc: @JnHs

PR #308, opened 2018-02-18
Update dialogs.md

c16u:
Fixing bug where the image was in line with text. It's now right under the paragraph alike to all the other images.


mikematteson:
@mijacobs  : PR **passed** review checklist. Please evaluate proposed change further and consider for merging. Note use of BR tag instead of extra line break to place image below text.

mijacobs (approved):

PR #312, opened 2018-02-23
Update use-mat.md

wstaelens:
info about neutralresource language that needs to be set

### DIFF
--- a/windows-apps-src/design/globalizing/use-mat.md
+++ b/windows-apps-src/design/globalizing/use-mat.md
@@ -1,6 +1,6 @@
----
+ ---
 author: stevewhims
-Description: The Multilingual App Toolkit (MAT) 4.0 integrates with Microsoft Visual Studio 2017 to provide UWP apps with translation support, translation file management, and editor tools.
+Description: The Multilingual App Toolkit (MAT) 4.0 integrates with Microsoft Visual Studio 2017 to provide UWP apps with translation support, translation file management, and editor tools. 
 title: Use the Multilingual App Toolkit
 template: detail.hbs
 ms.author: stwhi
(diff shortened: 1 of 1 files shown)


mikematteson:
@stevewhims : PR **passed** the acceptance checklist. Please evaluate further. Note also that a few adjustments to the submission would align it better with Microsoft style. Specifically:
- Avoid ellipses. (Lines 17, 60)
- Format names of UI elements in bold (e.g., "Properties", "Assembly Information").
- Use numbered lists, not bulleted lists, for sequential steps.
- Make items in lists (numbered or bulleted) grammatically and logically parallel.

Thanks!

stevewhims (commented):
Hi, @wstaelens, thanks for your pull request! The info you propose would be great for a topic about .NET, but not for this topic and the doc set that it's in (which is explicitly and exclusively scoped down to UWP apps). This topic has a link to the MAT download (https://marketplace.visualstudio.com/items?itemName=MultilingualAppToolkit.MultilingualAppToolkit-18308), and the MAT download page makes it clear what development technologies are supported. This topic also says "Before you can use the MAT effectively, your app needs to be localizable", and it links to a topic showing how to do that, the first step of which shows how to set your app's default language, which is what you do for a UWP app. For a UWP app there's no need to set the neutral language in order to overcome any MAT issues. So, am I right in thinking that you've been using the MAT with a .NET app, you ran into an issue, you fixed it using the steps you describe, and then you felt we should augment this topic with that info? If so then it's probably better to propose your change in a .NET topic, perhaps something like https://docs.microsoft.com/en-us/dotnet/framework/tools/.

wstaelens:
@stevewhims just wanted to share info about MAT here, feel free to change/edit/post/delete when/where needed.

stevewhims:
Great, well I'll pass this info along to the .NET content folks and also to the MAT development team. Once it's in the right hands, they'll be best situated to make the determination as to how to use the info. Again, many thanks for your contribution! And I'll close this PR once I've safely copied the info out of it.

PR #351, opened 2018-03-19
Replace unsuported markdown bits

binaryfunt:
e.g. :heavy_check_mark: does not format correctly in the docs website. Replaced with bullets


TylerMSFT:
Hi - this is a temporary issue and they will start rendering correctly again soon. Let's leave them as they are for now.

PR #427, opened 2018-04-09
Add C++/WinRT cross-link to C++/CX migration page

citelao:


mikematteson:
@stevewhims : PR **passed** [PR acceptance checklist](https://review.docs.microsoft.com/en-us/windows-authoring-guide/managing-contributions/editorial-checklist?branch=master). Please evaluate further and approve/merge if appropriate. Thanks!

stevewhims:
This topic **is a topic in the C++/WinRT section of the docs**. It's not a topic in the C++/CX docset, hence I included a link to the C++/CX docset. Is the intent of this PR to solve the issue of folks having dropped into this topic, and finding themselves in need of an "Intro to C++/WinRT" topic? If so, then solving that problem properly would involve replacing every occurrence of the string "C++/WinRT" in the C++/WinRT docs with a hyperlink to the "Intro to C++/WinRT" topic. Which feels very wrong to me. Hence, it feels wrong in this special case. Once a reader lands on this topic, or any of its siblings in the C++/WinRT docs, the table of contents in the browser shows them what topics are available, including the Intro topic. *That* feels like the solution. Whenever there's a mention of "C++/WinRT" *outside of* the context of the C++/WinRT docs, then we'll use that mention to link to the index page. But here we're already in there, so no need for a link IMO.

citelao:
@stevewhims I appreciate that reasoning. I added the crosslink for exactly the reason you mentioned.

I'm used to following the Wikipedia style for this, where the first instance of key words should be linked, no matter what. I think the table of contents is a good mitigator, but I think that cross-links are generally helpful. Of course, that's a decision for y'all on the content teams :)

I'm happy with a decision either way. Thanks!

stevewhims:
I see! I'll probably make the change you suggest to every first occurrence of "C++/WinRT" on every topic within the docset except the Intro topic itself. A lot of our topics link to themselves, which is a mistake IMO (since clicking the link is a confusing no-op for the reader), and this felt close in spirit to that, but I can see that it isn't. :) Oh, and I won't make the change to this particular topic, as I've now renamed the file so this PR won't succeed in any case. Please keep the suggestions coming, but if you could add your reasoning (or even email me directly), that'll make the intent clearer for me. Thanks! ~Steve

citelao:
Thank you for listening! I'm happy to keep sending feedback 😁.

stevewhims:
Great, thanks, Ben. I'll close this one, then. :)

PR #428, opened 2018-04-10
Add install of package coremltool

b02902131:
```
$ pip install coremltool
```
The command will fail because there is no suitable wheel package on pypi for windows.
```
$ pip install git+https://github.com/apple/coremltools
```
This command works.

Add this into this great article to make it even better.
In case there will still be someone who encounter the same problem.


mikematteson:
@serenaz: PR **passed** acceptance checklist. Please evaluate further and approve/merge if appropriate. Thanks!

serenaz (approved):
Thanks for the PR :)

PR #486, opened 2018-04-23
Update WDP API docs to reflect latest 1805 endpoints

WilliamsJason:
This PR should not be completed until the 1805 Xbox recovery ships. If there is a better holding branch to make the PR into, that would be awesome.


mikematteson:
Email sent to @WilliamsJason  and @mtoepke about alternative handling of this submission.

PR #500, opened 2018-04-28
Update sensor-orientation.md

Cloud-Writer:
From content request 80831 filed by Rob Caplan and approved by John Hornick: The columns in the current version of the article correctly describe landscape-first devices, but there is no information about what portrait-first devices should see, so that needs to be added.


Kellylorenebaker:
@GrantMeStrength : PR passes checklist review.

PR #549, opened 2018-05-10
Update icons.md

c16u:
Replaced two hyphens with the em dash.


mikematteson:
@mijacobs  : PR **passed** acceptance checklist. Please evaluate further, and note the following:
- Double hyphen replaced with em dash; better to use HTML entity code?

mijacobs (approved):

PR #568, opened 2018-05-14
fix typo (For for)

TylerMSFT:
Fixing remote since I'm not at work


mikematteson (approved):
@TylerMSFT : Looks good!

PR #658, opened 2018-06-08
Improve clarity of operation limits

Diego-Perez-Botero:
This article has conflicting guidance regarding operation limits. It also mentions System Event Broker, which is an internal/undocumented component, so it doesn't really add value to the guidance being given.

My team owns the BackgroundTransfer component, BTW ;)


mikematteson:
@stevewhims : PR **passed** review checklist. Please evaluate the proposed changes further. Thanks!

stevewhims (approved):
Looks great. Thanks, Diego!

PR #664, opened 2018-06-09
Fix implement typo

reepje123:


mikematteson (approved):
@reepje123 : Thanks for taking time to submit this fix. Much appreciated!

Cc: @anbare , @mijacobs

PR #725, opened 2018-06-28
Incorrect variable type used in screen capture sample.

DominicMaas:
A previous update changed the `_lastSize` variable to a `SizeInt32`, but never updated the `ResetFramePool()` method to make use of the new type. Updated to fix errors.

Not entirely sure why the 'See also' line claims it was updated.

### DIFF
--- a/windows-apps-src/audio-video-camera/screen-capture.md
+++ b/windows-apps-src/audio-video-camera/screen-capture.md
@@ -289,7 +289,7 @@ namespace CaptureSamples
             }
         } 
  
-        private void ResetFramePool(Vector2 size, bool recreateDevice) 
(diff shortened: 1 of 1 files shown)


mikematteson:
@eliotcowley : PR **passed** acceptance checklist. Please evaluate further and approve/merge if appropriate.

@DominicMaas: The extra diff on the "See also" section is probably just in white space--a difference in line or file terminators or something. Git should sort it out when we merge.

eliotcowley:
Thanks, @DominicMaas! I actually have an overhaul for this code snippet coming soon which fixes this, but good to have this fix in the meantime. Approved!

eliotcowley (approved):

PR #780, opened 2018-07-13
Improve wording of AttachAsync guidance

Diego-Perez-Botero:
Mention of "events" is confusing. Also, UploadOperations are subject to the same guidance. Most importantly, AttachAsync is something that should happen during app launch - it doesn't make sense to do it every time that a transfer is being started (like the current docs seem to suggest).


mikematteson:
@stevewhims : PR **passed** acceptance checklist. Please review further for technical accuracy, language, etc. Thanks!

stevewhims (approved):
Looks great; I just made one philosophical tweak. :) Thanks, Diego!

PR #1003, opened 2018-09-20
Rename Shim to PSF/Fixup

ArianGhotbiMSFT:
We are officially moving away from using the term Shim to a more appropriate term representing the technology.

The launchers are renamed to PSFLauncher.
Every runtime fix or any other fix, will now be named as Fixup.


normesta (changes requested):
It looks like the overview content is being added back to the article in this PR. I had removed this material earlier and instead point to a separate article with that overview. Can you please remove the addition of the overview material. I suspect its just because your branch hadn't pulled down the latest before pushing.

normesta (approved):

normesta:
Thank you Arian!

PR #1028, opened 2018-10-01
Improve MainPage.xaml.cs ( media playback with MediaSource )

touseefbsb:
1. code indentation improved along with some modern code practices like lambda expressions. 
2. Added foreach loop to add filetype filters to reduce code and make it simpler. 
3. PlayMediaSource should return a Task instead of a void because it is good practice that a async method should not return void but a event handler can return void even with async, so we can make 'PlayMediaSourceButton_Click' async and returning void instead.
4. Using new language features for null checking 'if (!(file is null))'
5. Added nullchecks for file and mediaPlaybackItem in method 'PlayMediaPlaybackItem'
6. new language features for string formating e.g :  '$"Track {index}"'


mikematteson:
@drewbatgit : PR **passed** review checklist. Please evaluate the proposed changes further. Thanks!

drewbatgit (approved):

PR #1034, opened 2018-10-02
Update app-icons-and-logos.md

kyleburnsdev:
Corrected spelling error


mikematteson:
@kyleburnsdev : Thanks for submitting this correction! As it happens, though, another contributor put in the *same fix* just a couple of hours before yours. Since that change is already in the pipeline, then (should be live in a couple of days), I'll close this PR. But we're grateful for the help and hope you won't hesitate to suggest changes whenever you see such issues in the future.

Cc: @mijacobs

PR #1067, opened 2018-10-07
Optimized headings

skycommand:
1. Promoted all headings to level 2 because headings that were most relevant to the topic of this article (i.e., creating, bundling, extracting, encrypting and decrypting) were all hidden under a level 2 heading called "Using MakeAppx.exe", which is a re-hash of the article title. Result: They are not shown on the table of contents. With my change, the reader should be able jump straight to the desired topic.
2. Converted imperative verbs in headings into declarative gerunds. The former was too forceful.


mikematteson:
@laurenhughes : PR **passed** checklist review. Please review for technical accuracy, etc., and approve/merge if appropriate.

> NOTE  There's no longer a firm MS style policy approving or disapproving the use of gerunds in headings. Consistency, within the topic and among topics in the same section, is the more important consideration. The imperative (or *bare infinitive*) form appears to be somewhat favored in the **Packaging apps** section overall, but @laurenhughes it's your call as content owner.

skycommand:
Huh! Microsoft is no longer mandated to use gerunds! That's a change after 30 years or so, isn't it?

Well, I don't think my change number 2 is something we'd want to argue for or against. So, please feel free to cherry-pick, @laurenhughes.

mijacobs (changes requested):
It looks like there are merge conflicts--can you rebase?

skycommand:
Stale

PR #1099, opened 2018-10-23
Remove errant mdash in C++/WinRT intro

plieblang:
Instead of appearing as a dash, it appeared literally as "&mdash", but since it's at the end of a sentence, it should be removed.


mikematteson (approved):
@plieblang : Thanks for spotting this and taking time to submit a fix. We appreciate the help!

Cc: @stevewhims

stevewhims:
Yes, thanks, @plieblang. FYI, a semi-colon at the end would have made that valid markdown (&mdash;). I forgot to tidy that up when I refactored the sentence. Thanks for fixing it! :) ~Steve

PR #1110, opened 2018-10-26
typo: Double word "specifying"

nschonni:


mikematteson:
@KevinAsgari : Is this in your content area? The PR change itself **passes** checklist review, but note that the topic has some other, general issues with metadata (ms.author) and miscoded headings. Fix along with this PR, or address independently? Please review and handle as appropriate. Thanks!

KevinAsgari:
Not specifically my area, but this pr should not be blocked, as the existing issues are already present.

KevinAsgari (approved):

PR #1143, opened 2018-10-28
chore: Remove merge artifact file

nschonni:
Looks like something from an old process


mikematteson:
@mijacobs : Okay to remove old branch-tracking test file?

mijacobs:
We haven't done anything with this file lately, but we're planning on using it in the future--thank you!

PR #1215, opened 2018-11-14
Update using-the-xaml-hosting-api.md

marb2000:
We need to collect feedback given this is in preview and we believe this is a good place for it.


mcleanbyron (approved):

mcleanbyron:
Thanks for contributing to these docs.

PR #1295, opened 2018-12-07
Stated dimensions

Heidelberger:
I added 1 line to explicitly state that the listed dimensions are width x height.

The included images are confusing and do not sufficiently communicate dimensions.  For example, the first image states 100 x 600 and shows a landscape-oriented screen. That implies that the dimensions are height x width (as the image is wider than it is high), however a bright blue box inside the image is portrait-oriented. The listed dimensions in fact refer only to that section. That's too many hurdles for the reader to jump. Explicitly stating width x height avoids possible confusion.

### DIFF
--- a/windows-apps-src/monetize/supported-ad-sizes-for-banner-ads.md
+++ b/windows-apps-src/monetize/supported-ad-sizes-for-banner-ads.md
@@ -9,9 +9,9 @@ ms.localizationpriority: medium
 ---
 # Supported banner ad sizes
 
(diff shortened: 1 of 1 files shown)


Kellylorenebaker:
@Xansky : PR **passed** the acceptance checklist. Please evaluate further and approve/merge if appropriate. Thanks!

mhopkins-msft (approved):

mhopkins-msft:
Thanks for the update!

PR #1329, opened 2018-12-20
Get example code to compile

MikeHillberg:
Code for static function x:Bind wasn't compiling.

### DIFF
--- a/windows-apps-src/data-binding/function-bindings.md
+++ b/windows-apps-src/data-binding/function-bindings.md
@@ -62,20 +62,17 @@ Static functions can be specified using XMLNamespace:ClassName.MethodName syntax
 <Page 
      xmlns:local="using:MyPage">
      ...
-    <Grid x:Name="myGrid" Background="Black" >
-        <TextBlock Foreground="{x:Bind local:MyColorHelpers.GenerateAppropriateForeground(myGrid.Background)}"
-                   Text="Hello World!" />
-    </Grid>
+    <StackPanel>
+        <TextBlock x:Name="BigTextBlock" FontSize="20" Text="Big text" />
+        <TextBlock FontSize="{x:Bind local:MyHelpers.Half(BigTextBlock.FontSize)}" 
+                   Text="Small text" />
+    </StackPanel>
 </Page>
 ```
 ```csharp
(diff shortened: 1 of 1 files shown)


windows-apps-src/data-binding/function-bindings.md
```diff
@@ -62,18 +62,19 @@ Static functions can be specified using XMLNamespace:ClassName.MethodName syntax
...
 ```csharp
-public class MyPage : Page
+static public class MyColorHelpers
 {
-    public static GenerateAppropriateForeground(SolidColorBrush background)
+    public static SolidColorBrush GenerateAppropriateForeground(Brush background)
     {
         //Implement static function
-        ...
+        return background as SolidColorBrush;
```
oldnewthing: This is the worst possible choice for foreground, hardly an "appropriate" foreground. Consider this alternative:

    public static SolidColorBrush GenerateAppropriateForeground(Brush background)
    {
        // The only type of brush we support is solid color brushes.
        var backgroundColor = ((SolidColorBrush)background).Color;
        // Formula taken from https://www.w3.org/TR/AERT/#color-contrast
        if (backgroundColor.R * 299 + backgroundColor.G * 587 * backgroundColor.B * 114 > 500 * 255)
        {
            return new SolidColorBrush(Colors.Black);
        }
        else
        {
            return new SolidColorBrush(Colors.White);
        }
    }

Or we can avoid the problem of explaining color space theory by choosing a simpler example.

    <TextBlock x:Name="BigTextBlock" FontSize="20" Text="Big text" />
    <TextBlock FontSize="{x:Bind local:MyHelpers.Half(BigTextBlock.FontSize)}" Text="Small text" />

    static public class MyHelpers
    {
        public static double Half(double value) => value / 2.0;
    }

MikeHillberg: Yes, I was trying to avoid color math.  I like the FontSize idea.

garycentric:
@mcleanbyron  : The PR **passed** the acceptance checklist. Please evaluate further and approve/complete the PR if appropriate. Thanks!

mcleanbyron (approved):
Thanks for catching the original problem and providing a better example!

PR #1417, opened 2019-01-25
Add note about 1809 and point to workaround

mrlacey:
Added note that this does not work on machines running 1809 and there are no plans to fix it. As per 
https://github.com/MicrosoftDocs/windows-uwp/issues/1389#issuecomment-457346505

### DIFF
--- a/windows-apps-src/monetize/launch-feedback-hub-from-your-app.md
+++ b/windows-apps-src/monetize/launch-feedback-hub-from-your-app.md
@@ -14,6 +14,9 @@ You can encourage your customers to leave feedback by adding a control (such as 
 To launch Feedback Hub from your app, use an API that is provided by the [Microsoft Store Services SDK](http://aka.ms/store-em-sdk). We recommend that you use this API to launch Feedback Hub from a UI element in your app that follows our design guidelines.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


garycentric:
@Xansky  @mrlacey :
PR **passed** acceptance checklist with the following (non-blocking) warnings. Please review further for technical accuracy, language, etc.

- :warning: References to a version of Windows by 4-digit year-month should follow this convention: "Windows 10, version 1809".
- :warning: Also, avoid blind links. Consider replacing "...no plans to fix it but [there is a workaround](https://stackoverflow.com/questions/53378611/storeservicesfeedbacklauncher-is-not-working/54299697#54299697)" with "...no plans to fix it, but there is a workaround; for more information, see  [StoreServicesFeedbackLauncher is not working](https://stackoverflow.com/questions/53378611/storeservicesfeedbacklauncher-is-not-working/54299697#54299697) on Stack Overflow.

mrlacey:
I'm holding off making these changes as I've heard a fix may be coming and so the warning/workaround will be unnecessary

mrlacey:
A new version of the SDK that addresses the issue was released last month. There is no need for this change now.

Se updated note at https://stackoverflow.com/questions/53378611/storeservicesfeedbacklauncher-is-not-working/54299697#54299697

PR #1513, opened 2019-03-02
Fix issues and inconsistencies in ICommand example

mrlacey:
Changes to address probelms with the code and also make the VM simpler so, hopefully, the important bits are more obvious:
- Remove invalid "Tooltip".
- Remove public fields from the VM.
- Bind to VM properties, rather than fields.
- Switched to auo-properties in the VM as fields not needed.
- Removed indication that VM implemented `INotifyPropertyChanged` as it didn't which meant the code was invalid, but it wasn't needed.
- Fixed bug where items may be incorrectly named when moving from right to left, so both list always show item 1 at the top and numbers increment going down.
- Simplify list initialization.
- Changed appropriate commands to use OneTime binding rather than OneWay as the binding is never changed.
- Fixed incorrect comment in the converter.

### DIFF
--- a/windows-apps-src/design/controls-and-patterns/commanding.md
+++ b/windows-apps-src/design/controls-and-patterns/commanding.md
@@ -612,10 +612,10 @@ We use two [ListViews](https://docs.microsoft.com/uwp/api/windows.ui.xaml.contro
             <StackPanel Grid.Row="1">
                 <FontIcon FontFamily="{StaticResource SymbolThemeFontFamily}" 
                           FontSize="40" Glyph="&#xE893;" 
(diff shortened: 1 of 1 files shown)


Kellylorenebaker:
@Karl-Bridge-Microsoft : PR **passed** the acceptance checklist. Please evaluate further and approve/merge if appropriate. Thanks!

Karl-Bridge-Microsoft:
@mrlacey : thanks for the suggested code revisions! Unfortunately, the sample no longer builds cleanly, so I have to spend some time looking into that before accepting your PR.

Karl-Bridge-Microsoft:
Thanks again @mrlacey, I've updated the sample with your suggestions, made a couple of other small tweaks, and confirmed all is working. I've pushed the updates in another PR, so I'll close this issue. Please re-open with any other related concerns.

PR #1698, opened 2019-05-23
Fix a few broken or obsolete links (.../desktop-to-uwp-distribute)

metathinker:
It seems the docs.microsoft.com publishing process is somewhat inconsistent in whether or not it auto-removes the .md extensions from linked page URLs when turning the markdown files into HTML webpages. This PR fixes a couple broken links resulting from that inconsistency.

I also updated a link that still works, but redirects.


garycentric:
PRRT: review required.

garycentric:
@metathinker, @mcleanbyron : This PR replaces broken links with complete, working URLs, but simply removing ".md" would make these affected links functional while also maintaining the relative paths to the pages. So, removing ".md" seems like a better solution.

Of course, I would keep the fix for the link that pointed to Technet.

Otherwise, this PR passes the acceptance checklist. @mcleanbyron, please evaluate further and approve/complete the PR if appropriate.

mcleanbyron (approved):

PR #1699, opened 2019-05-24
Code namespaces that map to the default XAML namespace

danzil:
Adding the list of code namespaces that map to the default XAML namespace.


mijacobs:
The private version of this PR was approved.

PR #1703, opened 2019-05-28
Change Items.Add to Items.Append

nxtn:
[ListView.Items](https://docs.microsoft.com/uwp/api/windows.ui.xaml.controls.itemscontrol.items#Windows_UI_Xaml_Controls_ItemsControl_Items) is a [ItemCollection](https://docs.microsoft.com/uwp/api/windows.ui.xaml.controls.itemcollection).

### DIFF
--- a/windows-apps-src/design/controls-and-patterns/listview-and-gridview.md
+++ b/windows-apps-src/design/controls-and-patterns/listview-and-gridview.md
@@ -80,11 +80,11 @@ Here's the list view created in code. The resulting list is the same as the one 
 ```csharp
 // Create a new ListView and add content. 
 ListView listView1 = new ListView(); 
-listView1.Items.Add("Item 1"); 
-listView1.Items.Add("Item 2"); 
-listView1.Items.Add("Item 3"); 
-listView1.Items.Add("Item 4"); 
-listView1.Items.Add("Item 5");
+listView1.Items.Append("Item 1"); 
+listView1.Items.Append("Item 2"); 
+listView1.Items.Append("Item 3"); 
+listView1.Items.Append("Item 4"); 
+listView1.Items.Append("Item 5");
  
 // Add the ListView to a parent container in the visual tree. 
 stackPanel1.Children.Add(listView1); 


garycentric:
@jwmsft: The PR **passed** the acceptance checklist. Please evaluate further and approve/complete the PR if appropriate. Thanks!

windows-apps-src/design/controls-and-patterns/listview-and-gridview.md
```diff
@@ -80,11 +80,11 @@ Here's the list view created in code. The resulting list is the same as the one
 ```csharp
 // Create a new ListView and add content. 
 ListView listView1 = new ListView(); 
-listView1.Items.Add("Item 1"); 
-listView1.Items.Add("Item 2"); 
-listView1.Items.Add("Item 3"); 
-listView1.Items.Add("Item 4"); 
-listView1.Items.Add("Item 5");
+listView1.Items.Append("Item 1");
```
MikeHillberg, line 83: Add is correct. 

Append is an extension method from the System.Linq.Enumerable class. It returns a new enumerable with the item appended to the current enumeration, but it doesn't modify the original list (doesn't modify listView1.Items). The purpose of this example is to call ICollection<object>.Add.

MikeHillberg (changes requested):

windows-apps-src/design/controls-and-patterns/listview-and-gridview.md
nxtn: [ItemCollection.Append](https://docs.microsoft.com/uwp/api/windows.ui.xaml.controls.itemcollection.append#Windows_UI_Xaml_Controls_ItemCollection_Append_System_Object_)

[ItemCollection](https://docs.microsoft.com/uwp/api/windows.ui.xaml.controls.itemcollection) doesn't implement any interface.

MikeHillberg: At the WinRT level, ItemCollection implements IVector<IInspectable>, which does have an Append method. But in the projection to C#, IVector<T> projects as IList<T>, which has an Add method instead (from ICollection<T>). So the members of ItemCollection depend on what language you're using.

I think the issue here is that the [ItemCollection](https://docs.microsoft.com/uwp/api/windows.ui.xaml.controls.itemcollection) pages aren't correct.

nxtn: Thanks. I was using C++ so I saw the `Append` method instead of `Add`.

MikeHillberg:
I opened a separate issue on ItemCollection class page to update/explain the different language projections.

PR #1800, opened 2019-07-15
Update documentation for 6.0.0-preview6.3

ocalvo:
In preview6.3 we introduce a native xaml application object that needs to be registered.


PRMerger20:
@ocalvo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger19:
@ocalvo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Per internal discussion with Oscar, this workaround is no longer necessary now that the preview6.4 NuGet packages for WPF and Windows Forms are available. I updated the current aritcle and related configuration guidance in [this article](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/xaml-islands#configure-your-project-to-use-xaml-islands) with the latest info to use the preview6.4 packages. This also means that the maxversiontested element is no longer necessary in these projects. Closing this PR because the doc updates have been addressed separately.

PR #1908, opened 2019-09-07
Scope shoud not limited to the "desktop" developer

hanrai:
I think this document is suitable for all Windows developers, not only for desktop developer. Because google bring me to the UWP section first. So I missed this important document. I spend two weeks to find this document out.
And I think, if you add a link of this document to all platform's overview section, it will help developers like me.( I was a Qt developer, I didn't know the differences between .NET, WinRT, WPF. So I mixed them all together in my mind.)

### DIFF
--- a/hub/apps/desktop/choose-your-platform.md
+++ b/hub/apps/desktop/choose-your-platform.md
@@ -11,7 +11,7 @@ ms.localizationpriority: medium
 
 # Choose your app platform
 
-When you want to create a new desktop application for Windows PCs, the first decision you make is which application platform to use. Windows provides four main application platforms, each with different strengths:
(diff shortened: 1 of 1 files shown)


PRMerger14:
@hanrai : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for your feedback and suggested changes to this article. We have decided to not accept these changes at this time, because this page and the related articles in this section are intentionally scoped to apps specifically written for desktop PCs (in contrast to apps for other Windows platforms such as IoT, HoloLens, etc.). You make a very good point about our other related docs (for WPF, Windows Forms, etc.) needing to tell a consistent story and point to this article for a comprehensive overview of all the available desktop platforms, and those changes are already in progress.

If you have further feedback or concerns, please feel free to reopen this with more details.

PR #1987, opened 2019-10-05
Fixed typo.

tim-weis:
Removed repetition in *"Then use **those the** stub type definitions ..."* to get *"Then use **those** stub type definitions ..."*.


PRMerger16:
@tim-weis : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #2020, opened 2019-10-17
Improvements to the custom xaml winrt/c++ control documentation

gbaychev:
I've added two notes to windows-apps-src\cpp-and-winrt-apis\xaml-cust-ctrl.md:

- a note about the fact that you need to download winrt/cpp extensions
if you use vs2019, otherwise you cannot create the tutorial project properly

- a note signifying that you need to be careful when creating the Themes
folder: if you simply right click your project -> add -> new filter, your folder won't be created. I've ran into this trap, because this menu item behaves/looks very similarly, but differently when you are working with C# projects. If you place your Generics.xaml in the wrong place, your custom control won't be displayed and you are going to have bad time.

### DIFF
--- a/windows-apps-src/cpp-and-winrt-apis/xaml-cust-ctrl.md
+++ b/windows-apps-src/cpp-and-winrt-apis/xaml-cust-ctrl.md
@@ -18,6 +18,9 @@ One of the most powerful features of the Universal Windows Platform (UWP) is the
 ## Create a Blank App (BgLabelControlApp)
 Begin by creating a new project in Microsoft Visual Studio. Create a **Blank App (C++/WinRT)** project, and name it *BgLabelControlApp*. In a later section of this topic, you'll be directed to build your project (don't build until then).
 
+> [!NOTE]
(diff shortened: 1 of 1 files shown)


PRMerger8:
@gbaychev : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):

PRMerger18:
@stevewhims : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Thanks for the contribution, @gbaychev. I've made a couple of changes. First, I've used my standard wording for referring folks back to the standard guidance on setting up their dev env for cppwinrt. Second, I've made it even more clear than before that we're asking readers to create a folder. If we can keep readers on the right track, then we needn't document what to do to get back on that track.

Thanks again!
Steve

gbaychev:
Thank you for accepting the PR, despite my clumsy formulations. It makes much more sense the way you've changed it. :)

PR #2068, opened 2019-11-06
Code-blocking time zones list

yoichinak:


PRMerger18:
@yoichinak : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys:
@yoichinak 

Thanks for taking the time to submit this PR. Can you explain a little more about what the utility of converting the time zone list from a table into a code block would be? Perhaps there's a workflow that we didn't think of when we created this table.

Thanks,
Shawn

yoichinak:
Thank you for your attention.
Before my PR, the time zone list looked like a simple text. To improve readability with small changes, I tried code-bolocking.
Now, it was table-made by https://github.com/MicrosoftDocs/windows-uwp/commit/072505ffb33c56c7de29d331164a67f1788b0b0c#diff-81e511069c8d9359b8e97402c5313dc5 . That is the best solution :)

hickeys:
Thanks @yoichinak. By the time I looked at it, it was already a table. Sounds like a code block would have been an upgrade from no formatting at all, and it's on me for waiting two months to review your PR. Thanks for your contribution, and I'll try to look at it more quickly next time :)

PR #2071, opened 2019-11-07
Correct incorrect use of UX & supported features

mrlacey:
UI and UX are not the same thing. UX is more that just the interface or presentation layer. Don't use UX when mean UI.

Dense (compact) UI is supported in UWP as of 18362(1903) - https://docs.microsoft.com/en-us/windows/uwp/whats-new/windows-10-build-18362#design--ui
There is already a disclaimer that "Some UWP APIs require specific minimum versions of Windows 10" so I don't think it's necessary to call out the platform support specific to Dense UI.


PRMerger17:
@mrlacey : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron (approved):

PR #2150, opened 2019-12-30
Update display title to match heading

miniksa:
The display title on this page does not match the heading on the page. I was confused looking through my open tabs to find one that started with "Strong and weak" and didn't find one. Turns out, the header doesn't match the metadata that makes the tab title.

This resolves that minor annoyance by making the metadata match the rest of the page.

I chose to bias toward "strong and weak" (instead of changing the heading to "weak" alone) because the heading says that and so does the TOC link. Two out of three votes used both, so I took that one.


stevewhims:
Thanks, @miniksa!

PR #2264, opened 2020-03-05
move-to-winrt-from-cx.md: Add code for array delay init

fghzxm:
Currently, /windows-apps-src/cpp-and-winrt-apis/move-to-winrt-from-cx.md states that there is no automated way to construct an array of empty (delayed-initialized) references in C++/WinRT.  This commit provides a TMP-based solution to show that such objective is indeed accomplishable, even if the solution is ad-hoc and verbose.

Because such requirement is arguably rarely encountered in the real world, the updated article puts the code and its surrounding paragraphs inside a `[!NOTE]`.

Problems that this commit may have:
1. The implementation listed in this commit lacks the divide-and-conquer optimization that would reduce the time and space complexity from O(N^2) to O(Nlog N).  Since we do not expect the reader to implement the optimization by themselves, we may want to update this solution to use the optimized algo.
2. On the other hand, the code in this commit is already long and potentially daunting to read, and optimizing it would only make it worse.
3. As mentioned in the added paragraphs, the hand-rolled code in this commit will be superseded in the future when C++ gets a library function that does the equivalent of `std::vector::vector(size_t, const T&)` for `std::array`.

Possible alternatives to merging this commit:
1. Put the code somewhere else on the web and link to that page in this article.
2. Add the proposed code to C++/WinRT instead.

### DIFF
--- a/windows-apps-src/cpp-and-winrt-apis/move-to-winrt-from-cx.md
+++ b/windows-apps-src/cpp-and-winrt-apis/move-to-winrt-from-cx.md
@@ -304,7 +304,65 @@ C++ collection types use the default constructor, which can result in unintended
 | Array of empty references | `TextBox^ boxes[2];` | `// Creates 2 TextBox objects!`<br/>`TextBox boxes[2];` | `TextBox boxes[2] = { nullptr, nullptr };` |
 | Pair | `std::pair<TextBox^, String^> p;` | `// Creates a TextBox!`<br/>`std::pair<TextBox, String> p;` | `std::pair<TextBox, String> p{ nullptr, nullptr };` |
 
-There's no shorthand for creating an array of empty references. You have to repeat `nullptr` for each element in the array. If you have too few, then the extras will be default-constructed.
+There's no shorthand for creating an array of empty references. If you use aggregate initialization but provide too few `nullptr`s, then the extras will be default-constructed:
+
+```cppwinrt
+// Still creates 2 TextBox objects!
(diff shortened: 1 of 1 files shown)


PRMerger13:
@fghzxm : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger17:
@fghzxm : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Hi, @fghzxm, and thanks for taking the time to create this solution. I feel like this case is esoteric enough not to belong in the mainstream content. Would it be suitable for a gist on your GitHub account?

As for getting your solution integrated into C++/WinRT itself, I'd recommend that you pursue that as feeback/questions about the product itself (rather than a PR for the documentation). If you visit the target topic online and click **Send feedback about** \> **This product**, you'll be taken to [Microsoft Q&A](https://docs.microsoft.com/answers/topics/uwp.html?sort=active&filter=all), which is the right place to lobby for a change, a feature, or to make a suggestion.

I suggest we close the PR, though.

Thanks again!
Steve

fghzxm:
OK, I'll leave a comment on microsoft/cppwinrt.

stevewhims:
Thanks, @fghzxm. I've updated the topic to make it clearer that in C++/WinRT we can choose a **std::vector** in cases like this, for convenience (in fact, we can use any contiguous container). We don't have that option in C++/CX. Thanks for your feedback!

PR #2294, opened 2020-03-19
Fix typo in StateTriggers call-out

potto007:
Change "attacked" to "attached"

While more exciting, I don't think VisualStateGroups "attack" any children.


PRMerger6:
@potto007 : Thanks for your contribution! The author(s) have been notified to review your proposed change.

QuinnRadich:
Hah! Well, that's an amusing typo. I'm half-tempted to just leave it as it is...

Good catch.

QuinnRadich (approved):

PR #2301, opened 2020-03-21
Head first word for classic COM interfaces support

yangrongwei:


PRMerger7:
@yangrongwei : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
Hi, @yangrongwei, and thanks for proposing these changes. I feel like the changes proposed don't improve the intent nor flow of the topic. My idea here is to first explain what happens by default, and why. At that stage, I'd like the picture to be as simple and clear as possible and I feel like it might possibly be confusing, or too much info, to talk about the solution at the same time as talking about what happens when you don't use that solution. If we mention the header file where you suggest, and then show an example that doesn't use the header file, that feels potentially confusing, and like it's diluting, and redirecting, the flow of the topic (which is show problem, then show solution).

So I'll go ahead and close this PR, and I'll take a look at the second one.

Thanks again!
Steve

PR #2302, opened 2020-03-21
Head frist word for classic COM interfaces support

yangrongwei:


PRMerger10:
@yangrongwei : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims:
So, we need to include `unknwn.h` when we're using **winrt::implements** to implement a COM interface. And it's advisable to include it when we call **winrt::get_unknown**. However, in the code example where you suggest that we mention including `unknwn.h`, we're not doing either of those things, so we don't need to include it, nor mention it.

In the later code example we are calling **winrt::get_unknown**, so we should include `unknwn.h` there. I have actually gone ahead and added more info to the topic, so hopefully it'll be clearer.

I also feel like we don't need to emphasize "immediately after".

I'll go ahead and close this PR, but please leave feedback on any of the topics if they're wrong, or unclear.

Thanks!
Steve

PR #2311, opened 2020-03-23
Fixed inconsistency in tutorial runserver command 

misczak:
In the Hello World tutorial for Django, fixed an inconsistency in Step 17 where the instructions said to use 'python' to run the server, not 'python3' like the rest of the tutorial.


mattwojo:
Great catch. Thanks again @misczak

PR #2314, opened 2020-03-24
Update XCG screenshot

marcelwgn:
This PR updates the screenshot for the XAML Controls Gallery, which was outdated.


ktoliver:
@jwmsft Hi, it looks like you submitted the commit that created the original file. Could you take a look at this proposed update? Thanks.

jwmsft (commented):
@chingucoding , thanks for taking the time to submit this update, we appreciate it. However, in this case, I'm going to close this PR. We're actually in the process of replacing the XCG UI screenshot image with a logo image (because it's easier to keep up-to-date). Most articles were updated to use the logo image last week, and the remaining articles will be updated this week, and then the UI image will no longer be used.

marcelwgn:
Thanks for the response and the detailed explanation @jwmsft !

PR #2368, opened 2020-04-17
Update SetTitleBar documentation to point out issue with grid

marcelwgn:
When a grid has its background unset, it does not take part in hit testing. That's why you need to set the background, however that is not trivial. That's why we should point that out.

See this issue for more context:
https://github.com/microsoft/microsoft-ui-xaml/issues/1004


PRMerger6:
@chingucoding : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys (approved):

PR #2428, opened 2020-05-08
Fix JSON string to actually parse

tonyferrell:
I copied this example and was getting exceptions from the API. It appears the sub-object was wrapped in "'s when it should not be.


PRMerger14:
@tonyferrell : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys (approved):

hickeys:
#sign-off

PR #2491, opened 2020-05-26
TabItems is the Content property now

michael-hawker:
This was updated to not be needed: https://github.com/microsoft/microsoft-ui-xaml/pull/1845


hickeys:
#sign-off

PR #2513, opened 2020-06-08
Removed generated file reference 'app.g.cpp'

SkyeHoefling:
The `app.g.cpp` file is never generated and sample code projects do not include it. 

The sample code provided does not include it

https://github.com/microsoft/Xaml-Islands-Samples/blob/0d84252237132f0e02421c6b7b22edff7dd6a803/Standalone_Samples/CppWinRT_Desktop_Win32App/UWPApplication/App.cpp#L1-L17

### DIFF
--- a/hub/apps/desktop/modernize/host-custom-control-with-xaml-islands-cpp.md
+++ b/hub/apps/desktop/modernize/host-custom-control-with-xaml-islands-cpp.md
@@ -238,7 +238,6 @@ Next, revise the default **App** class in the **MyUWPApp** project to derive fro
         ```cpp
         #include "pch.h"
         #include "App.h"
-        #include "App.g.cpp"
         using namespace winrt;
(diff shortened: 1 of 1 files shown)


PRMerger9:
@ahoefling : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mcleanbyron:
Thanks for your feedback. In the context of this walkthrough article, which has you build all the projects from scratch, this is a necessary line of code. By default in new C++/WinRT UWP projects, the project-level **Optimized** property is set to **yes**, and this setting requires this line of code. I have added a note just after the [App.cpp code example](https://docs.microsoft.com/en-us/windows/apps/desktop/modernize/host-custom-control-with-xaml-islands-cpp#define-a-xamlapplication-class) in the article to clarify this detail.

PR #2531, opened 2020-06-18
Change Protocol Activation Parameter from Optional

wjwalcher:
I work at Microsoft (on the Windows Media Apps team - feel free to ping me on Teams 👍) and recently encountered an issue when trying to launch a video for editing with this URI scheme. I was only specifying the InputToken parameter here, since the others were indicated as optional, and my video would never open, I just get a black screen. 
Feel free to try it with the Windows Store build of Photos and you'll see this same behavior. I dug around in the code and it turns out that you do need to specify this parameter for the videoedit activation to work. As such, it should be changed from an optional parameter to a required one.


PRMerger16:
@wjwalcher : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hkhmsft:
Thanks for the contribution, @wjwalcher ! I've merged your changes and they should be live soon.

PR #2550, opened 2020-06-26
Update ARM64 SDK announce link

marswe:
The link for Build ARM64 apps with the SDK should point to the blogpost announcing the actual release rather than the blogpost announcing the preview.


ktoliver:
@msatranjr - Could you review the proposed update to the ARM docs landing page? Thanks.

mcleanbyron (approved):

PR #2886, opened 2020-12-26
Fix Segoe icon code typo

LaurentTreguier:
This looks like a small leftover from a copy-paste of another symbol above. Given that the symbols above and below `SerbianSpelling` are `F87C` and `F87E`, I tried using `F87D` in a FontIcon, and it seems to be the correct code.


PRMerger9:
@LaurentTreguier : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys (approved):

PR #2940, opened 2021-01-29
Fixing a missing `#endif`

LoadLibrary:
The application doesn't compile after copying and pasting the provided source code.
This change adds the missing `#endif` that allows the code to compile.


drewbatgit (approved):

drewbatgit:
Thanks for catching this!

PR #2998, opened 2021-03-05
Add Rust operator link and comments

camerondurham:
Added comments in front of `...` to make it easier to use the Microsoft Doc's great **Copy** feature without syntax errors.  This is a very small change but makes it easier to quickly and lazily copy/paste the codeblock into your local `.toml` or `.rs` file.

Added links to relevant docs about Rust's `?` operator.

This tutorial was overall very easy to follow and worked seamlessly for me. Thanks for the great job making it easy to get started with Rust on Windows!

### DIFF
--- a/hub/dev-environment/rust/rss-reader-rust-for-windows.md
+++ b/hub/dev-environment/rust/rss-reader-rust-for-windows.md
@@ -166,7 +166,7 @@ Because [**RetrieveFeedAsync**](/uwp/api/windows.web.syndication.syndicationclie
    ```rust
    // src\main.rs
    // ...
-   
+
    fn main() -> Result<()> {
        let uri = Uri::create_uri("https://blogs.windows.com/feed")?;
        let client = SyndicationClient::new()?;
--- a/hub/dev-environment/rust/rss-reader-rust-for-windows.md
+++ b/hub/dev-environment/rust/rss-reader-rust-for-windows.md
@@ -51,7 +51,7 @@ Now let's try out Rust for Windows by writing a simple app that downloads the ti
 
    ```toml
    # bindings\Cargo.toml
-   # ...
+   ...
 
    [dependencies]
    windows="0.3.1"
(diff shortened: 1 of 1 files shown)


PRMerger19:
@camerondurham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger6:
@camerondurham : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger10:
@stevewhims : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Hi, @camerondurham, and thanks for the PR! :) I've taken your suggestions, with the exception of commenting out the ellipses. Across my docs, I have a number of cases where I show a small piece of source code and then successively build that up over the course of a number of listings. In so doing, in each listing I show only the deltas, and I include enough context to make it clear where the reader should add the delta to their own project. I don't, of course, repeat the entire previous listing + delta. That would take up too much space, and would make it harder to identify the delta. Any "delta" listing isn't meant to compile on its own, so it's not meant to be copy-pasted in its entirety into an entire source code file. The reader can of course use the Copy affordance, if that's convenient. But in many cases it's easier just to select-copy the delta and paste that in to the right place.

Anyway, consequently, to remain consistent with the practice described above (which I use in several places in the C++/WinRT docs), I've set the ellipses back to their sans comment syntax form. Remember that those ellipses are not meant to be source code (which the comment might imply) any more than an ellipsis in a passage of English is meant to be English. In both cases (source code or human language), rather, they're meant to indicate that something has been intentionally elided (with enough context to be clear exactly *what* has been elided).

Thanks!
Steve

camerondurham:
Hi @stevewhims, 
Your reasoning completely makes sense. Focusing on the delta does make it easier to follow the tutorial.

Thank you for the explanation. I appreciate that you took the time to explain that. This will help me write better docs in the future!

stevewhims:
Awesome! Thanks for your contributions! :)

Issue #3045, opened 2021-04-01
vault.FindAllByResource(resourceName); throws exception if resource is not found. Example does not reflect this behavior


SvenLauterbach:
Hi,

The example from this page seems a little bit buggy.

```csharp
var credentialList = vault.FindAllByResource(resourceName);
```

According to the [documentation ](https://docs.microsoft.com/en-us/uwp/api/windows.security.credentials.passwordvault.findallbyresource?view=winrt-19041) `FindAllByResource` throws an exception:

> The resource to be searched for. If a match does not exist, an exception will throw.

This means, that the `else`block is never reached:

```csharp
var loginCredential = GetCredentialFromLocker();

if (loginCredential != null)
{
    // There is a credential stored in the locker.
    // Populate the Password property of the credential
    // for automatic login.
    loginCredential.RetrievePassword();
}
else
{
    // There is no credential stored in the locker.
    // Display UI to get user credentials.
    loginCredential = GetLoginCredentialUI();
}
```

The exception should be catched by `GetCredentialFromLocker` and than return `null`.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 350b7fbc-954c-be69-79b1-e0c121f16070
* Version Independent ID: bc24d4ba-c7fc-bd38-5455-37486de1af80
* Content: [Credential locker - UWP applications](https://docs.microsoft.com/en-us/windows/uwp/security/credential-locker)
* Content Source: [windows-apps-src/security/credential-locker.md](https://github.com/MicrosoftDocs/windows-uwp/blob/docs/windows-apps-src/security/credential-locker.md)
* Product: **uwp**
* Technology: **platform**
* GitHub Login: @hickeys
* Microsoft Alias: **hickeys**

hickeys:
Great catch. This was updated recently so I believe this issue has been resolved.

PR #3046, opened 2021-04-01
Fix credential locker example

SvenLauterbach:
Hi,

this PR addresses the issue #3045. It implements the proposed solution by catching the exception thrown by `vault.FindAllByResource` and returns null.

### DIFF
--- a/windows-apps-src/security/credential-locker.md
+++ b/windows-apps-src/security/credential-locker.md
@@ -86,7 +86,7 @@ private Windows.Security.Credentials.PasswordCredential GetCredentialFromLocker(
     {
         credentialList = vault.FindAllByResource("bunq");
     }
-    catch(Exception ex)
+    catch(Exception)
     {
         return null;
     }
--- a/windows-apps-src/security/credential-locker.md
+++ b/windows-apps-src/security/credential-locker.md
@@ -79,7 +79,18 @@ private Windows.Security.Credentials.PasswordCredential GetCredentialFromLocker(
     Windows.Security.Credentials.PasswordCredential credential = null;
 
     var vault = new Windows.Security.Credentials.PasswordVault();
-    var credentialList = vault.FindAllByResource(resourceName);
+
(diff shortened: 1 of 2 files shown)


PRMerger12:
@SvenLauterbach : Thanks for your contribution! The author(s) have been notified to review your proposed change.

SvenLauterbach:
Actually, it would be better to update the behavior of the `FindAllByResource` method, because not finding a requested resource is not something unexpected, but I didn't know where I could address this and I also think, this might be harder to fix than the example.

hickeys (changes requested):
I like this change in principal, but it needs two changes:

1. Pass `resourceName` to `vault.FindAllByResource` instead of `"bunq"`
2. Catch the specific exception you're looking for instead of catching all exceptions

Make those changes and this looks great!

SvenLauterbach:
The first one is easy, but the second suggestion is hard, since I don't know, which exception is thrown if there is no resource found. I event tried `ex.GetType().FullName` which yield `System.Exception`, so this seems to be the most specific exception.

The best I could do, to narrow the exception down to this specific exception, is:

```cs
catch(Exception ex) when (ex.Message.EndsWith("Cannot find credential in Vault"))
{
    return null;
}
```

Would this be ok?

PRMerger16:
@SvenLauterbach : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys:
Well that's unfortunate. I like the way you're thinking for ways to identify the exception without looking at its type, but I worry this won't work in regions where the exception text is in another language.

I'm going to do a little reading up on this. I'll get back to you if I find a good solution.

hickeys:
I haven't been able to find a better solution. Let's go back to your original proposal where simply caught `Exception`

PRMerger14:
@SvenLauterbach : Thanks for your contribution! The author(s) have been notified to review your proposed change.

hickeys (approved):

PR #3103, opened 2021-05-05
Update run.md

tomverhoeff:
Updated the known issues link to use the right label for PowerToys Run issues. It was still referring to PowerToys Launcher which is not used anymore and does not yield any results.


PRMerger18:
@tomverhoeff : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mattwojo:
Thanks @tomverhoeff ! Great catch!

PR #3125, opened 2021-05-18
Update attached-properties-overview.md

jaigak:
Fixes indentation for C#, VB and C++/CX. Also removes dev langs as they were not working correctly and just shows the examples in all the languages by default.

### DIFF
--- a/windows-apps-src/xaml-platform/attached-properties-overview.md
+++ b/windows-apps-src/xaml-platform/attached-properties-overview.md
@@ -6,11 +6,6 @@ ms.date: 02/08/2017
 ms.topic: article
 keywords: windows 10, uwp
 ms.localizationpriority: medium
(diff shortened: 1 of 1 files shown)


jwmsft (changes requested):
@Jaiganeshkumaran, thanks for helping us fix these code samples. The devlangs need to exactly match the code blocks in the doc, so the correct fix here is to update the code blocks from `cpp` to `cppcx`, and update the devlangs to:
```
dev_langs:
  - csharp
  - vb
  - cppwinrt
  - cppcx
```
Thanks

ktoliver:
It looks like the working branch that was used to create the pull request has been deleted, so a commit can't be added to the pull request. We'll close the PR. Please create a new pull request to update the article per the author's feedback. Thanks.

PR #3127, opened 2021-05-18
Fix ARM64 Project XML Syntax

michael-hawker:


PRMerger17:
@michael-hawker : Thanks for your contribution! The author(s) have been notified to review your proposed change.

michael-hawker:
Also at the start of this section: https://github.com/MicrosoftDocs/windows-uwp/blob/docs/windows-apps-src/cpp-and-winrt-apis/use-csharp-component-from-cpp-winrt.md#building-for-net-native

It seems 'recommended' should be 'required' currently, as there is this issue that I opened on the WinUI repo to get my scenario to work for User Controls in XAML: https://github.com/microsoft/microsoft-ui-xaml/issues/5040

stevewhims (approved):

PR #3165, opened 2021-06-08
Update navigation-history-and-backwards-navigation.md

martingbrown:
Change the code examples for handling mouse back and forward buttons to move back and forward if the event has not already been handled.

### DIFF
--- a/windows-apps-src/design/basics/navigation-history-and-backwards-navigation.md
+++ b/windows-apps-src/design/basics/navigation-history-and-backwards-navigation.md
@@ -472,12 +472,12 @@ private void CoreWindow_PointerPressed(CoreWindow sender, PointerEventArgs e)
 {
     // For this event, e.Handled arrives as 'true', so invert the value.
     if (e.CurrentPoint.Properties.IsXButton1Pressed
-        && e.Handled)
+        && !e.Handled)
     {
         e.Handled = !TryGoBack();
     }
     else if (e.CurrentPoint.Properties.IsXButton2Pressed
-            && e.Handled)
+            && !e.Handled)
     {
         e.Handled = !TryGoForward();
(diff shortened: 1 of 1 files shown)


ktoliver:
Hi @martingbrown - Thank you for your contribution. Would you take a moment to sign the Contributor License Agreement (CLA)? After the CLA is signed, someone can review your pull request. Thanks!

PRMerger15:
@martingbrown : Thanks for your contribution! The author(s) have been notified to review your proposed change.

martingbrown:
@ktoliver I've signed the CLA.

ktoliver:
@martingbrown Thanks!

@jwmsft  Could you take a look at the proposed update? Thanks.

jwmsft:
@martingbrown, thanks for submitting this update. This `CoreWindow_PointerPressed` handler is a strange case, because `e.Handeled` arrives as `true`, so it actually works correctly as-is. If you change the code to check that the event is _not_ already handled, `TryGoForward` and `TryGoBack` are never executed. That's why the check is reversed from what you would expect.

However, after testing it a bit, there doesn't seem to any negative side effects to not checking `e.Handled` at all, so that check could just be removed.

```
    if (e.CurrentPoint.Properties.IsXButton1Pressed)
    {
        TryGoBack();
    }
    else if (e.CurrentPoint.Properties.IsXButton2Pressed)
    {
        TryGoForward();
    }
```

If you'd like to update this PR with this change, I can approve it. Thanks.

PRMerger17:
@martingbrown : Thanks for your contribution! The author(s) have been notified to review your proposed change.

jwmsft (approved):

jwmsft:
#sign-off

PR #3320, opened 2021-08-26
Fix C# code sample typo

radujipa:
In C# the [IReadOnlyList<T> returned for args.Files](https://docs.microsoft.com/en-us/uwp/api/windows.applicationmodel.activation.fileactivatedeventargs.files?view=winrt-20348#Windows_ApplicationModel_Activation_FileActivatedEventArgs_Files) does not have a `.Size` property, but it does have a `.Count`. I am not sure about the other languages though I suspect it may be the same typo than needs verifying and correcting.


PRMerger16:
@radujipa : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3342, opened 2021-09-11
Update PowerToys docs

Jay-o-Way:
Mostly focussed on tables and kbd tweaks.
See https://github.com/microsoft/PowerToys/issues/13165
FYI @dedavis6797

### DIFF
--- a/hub/powertoys/administrator.md
+++ b/hub/powertoys/administrator.md
@@ -26,7 +26,7 @@ The simplest way to run an app or program in administrative mode is to right-cli
 
 Most apps do not need to run with elevated permission. A common scenario, however, for requiring administrator permission would be to run certain PowerShell commands or edit the registry.
 
-If you see this prompt (User Access Control prompt), the application is requesting administrator level elevated permission:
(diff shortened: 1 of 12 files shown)


PRMerger17:
@Jay-o-Way : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PRMerger12:
@mattwojo : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mattwojo:
Thanks for all of these updates @Jay-o-Way! One note - we generally try to use the language of "select", rather than "click" or "press" in our docs in an effort to be more inclusive of users who may be using adaptive devices, but I'm approving for a merge with the current terms pending a conversation in the PowerToys dev meeting to see how folks feel this should be handled moving forward (as I'm not confident that I understand the full use-case picture in regard to PT). Everything else looks great. Thanks again!

Jay-o-Way:
Oh I see. My thought was that "press" sounds more like an action and "select" sounds like "choose", which is not so much an action. Also used "press" in favor of "click" to include touch. But you're right, there are other input methods.

mattwojo:
Totally makes sense.

PR #3374, opened 2021-09-28
Update set-custom-permissions-for-account-users.md

nickrhegge:
New account level permission was added to enable ordering Xbox Dev Kits and this permission gates user level access.


PRMerger13:
@nickrhegge : Thanks for your contribution! The author(s) have been notified to review your proposed change.

PR #3379, opened 2021-10-01
Update asynchronous-programming-in-cpp-universal-windows-platform-app…

jaigak:
Changes "UWP" to "Windows Runtime" in some places where it isn't specific to UWP apps.


PRMerger10:
@Jaiganeshkumaran : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ShannonLeavitt:
#label:"aq-pr-triaged"

drewbatgit:
@stevewhims  This looks pretty straightforward to me, but I feel like you have a better grasp on the precise terminology we should use here. Will you take a look and approve or reject this PR?

stevewhims (approved):
Yes, I'm happy with this change. These *are* Windows Runtime types, although I'm not sure the distinction is important for this topic since I'm unaware that you can use C++/CX anywhere but a UWP app. But whether you can or can't, the change here doesn't break anything, so I think it'd be good to take.

PR #3438, opened 2021-10-27
Updated note on stopping haptics

KybeckMicrosoft:
Updating the note regarding pens automatically stopping haptics when the pen leaves range of the screen.  Previously stated that pens automatically do this, however, that is an OPTIONAL implementation detail for pens, so I am updating the note to properly reflect this.


PRMerger16:
@KybeckMicrosoft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #3511, opened 2021-11-23
Include keyboard accelerators in tooltips

mcooley:
The [keyboard accelerators documentation](https://docs.microsoft.com/en-us/windows/apps/design/input/keyboard-accelerators#tooltips) includes a section on tooltips, but the [tooltips documentation](https://docs.microsoft.com/en-us/windows/apps/design/controls/tooltips) doesn't say anything about keyboard accelerators. Add a suggestion that when you override the default behavior, you should probably include information about keyboard accelerators so that users can discover them.

Also, the first tooltip in the article ("Click to submit") doesn't follow our recommendation: "Don't make it obvious or just repeat what is already on the screen.". I changed this to "Create a new document", which is a more realistic example where a tooltip might be needed to provide supplemental information.

### DIFF
--- a/hub/apps/design/controls/tooltips.md
+++ b/hub/apps/design/controls/tooltips.md
@@ -85,18 +85,18 @@ A [ToolTip](/uwp/api/Windows.UI.Xaml.Controls.ToolTip) must be assigned to anoth
 In XAML, use the **ToolTipService.Tooltip** attached property to assign the ToolTip to an owner.
 
 ```xaml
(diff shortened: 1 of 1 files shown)


PRMerger20:
@mcooley : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

jwmsft (approved):

jwmsft:
#sign-off

PR #3543, opened 2021-12-05
Added C++ snippits to mica.md

Koyuki-of-SimpleIO:
Added C++ code snippits that are confirmed fully functional using VS 2022
Also added the header file includes

### DIFF
--- a/hub/apps/design/style/mica.md
+++ b/hub/apps/design/style/mica.md
@@ -437,6 +437,125 @@ private void NavigationViewControl_DisplayModeChanged(Microsoft.UI.Xaml.Controls
 }
 ```
 
+```cpp
+MainPage::MainPage()
+{
+	InitializeComponent();
+    auto titleBar = ApplicationView::GetForCurrentView()->TitleBar;
+
+    titleBar->ButtonBackgroundColor = Windows::UI::Colors::Transparent;
+    titleBar->ButtonInactiveBackgroundColor = Windows::UI::Colors::Transparent;
+
+    // Hide default title bar.
+    auto coreTitleBar = CoreApplication::GetCurrentView()->TitleBar;
+    coreTitleBar->ExtendViewIntoTitleBar = true;
+    UpdateTitleBarLayout(coreTitleBar);
+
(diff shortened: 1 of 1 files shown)


PRMerger9:
@Koyuki-of-SimpleIO : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

sylveon:
Why C++/CX instead of C++/WinRT?

Koyuki-of-SimpleIO:
> Why C++/CX instead of C++/WinRT?

Comfort mainly, have written in CX for a while. If you would like I can update the example to follow WinRT rules, but overall I really do prefer CX.

sylveon:
It should be in cppwinrt, as C++/CX is deprecated and cppwinrt is recommend for all new development, so docs shouldn't be recommending it.

Koyuki-of-SimpleIO:
Got it, will update to RT tonight.

hickeys (approved):
These look fine to me, but my c++ is rusty. I'm requesting a review from someone who knows the language much better than me.

stevewhims (changes requested):
It was discussed in the conversation thread that this code needs porting from C++/CX to C++/WinRT. The thing is, if there is only ONE flavor of C++ in a topic then that flavor needs to be C++/WinRT. There's already a C# version of the code, so the topic meets the min bar for code examples. If I get bandwidth I'll be happy to port that existing C# version to C++/WinRT (and in fact it's often easier to go from C# to C++/WinRT than from C++/CX to C++/WinRT). But I'd recommend closing this PR, unless we can get it changed to C++/WinRT.

stevewhims:
Oh, also the language slug needs a tweak. Since this is C++/CX (which we shouldn't publish) the slug for that is ```cppcx. But once it's ported to C++/WinRT, it'd be ```cppwinrt.

hickeys (approved):

stevewhims:
This one is very old now, and isn't being actively worked from either side. Let's close it. Thanks, all! :)

-Steve

PR #3557, opened 2021-12-10
Update resource guidance to match NuGet package readme

sylveon:
The readme was telling users to use a different way. Update the docs to also prefer this method.

### DIFF
--- a/hub/apps/winui/winui2/getting-started.md
+++ b/hub/apps/winui/winui2/getting-started.md
@@ -54,29 +54,27 @@ The library is available as a NuGet package that can be added to any new or exis
     </Application>
     ```
 
-    b. If you need more than one application resource, add the WinUI resources element `<XamlControlsResources` in a `<ResourceDictionary.MergedDictionaries>` as shown here:
+    b. If you have other resources, then we recommend you add those to `XamlControlsResources.MergedDictionaries`. This works with the platform's resource system to allow overrides of the `XamlControlsResources` resources.
 
     ``` XAML
     <Application
         x:Class="ExampleApp.App"
         xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+        xmlns:controls="using:Microsoft.UI.Xaml.Controls"
         RequestedTheme="Light">
 
         <Application.Resources>
-            <ResourceDictionary>
-                <ResourceDictionary.MergedDictionaries>
-                    <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
+            <controls:XamlControlsResources>
+                <controls:XamlControlsResources.MergedDictionaries>
                     <ResourceDictionary Source="/Styles/Styles.xaml"/>
-                </ResourceDictionary.MergedDictionaries>
-            </ResourceDictionary>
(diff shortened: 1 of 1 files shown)


PRMerger14:
@sylveon : Thanks for your contribution! The author(s) have been notified to review your proposed change.

ktoliver:
#label:"aq-pr-triaged"

Panda-Sharp:
I'd suggest to clarify once and for all what's the correct way, also would be nice to provide some samples

cc https://github.com/microsoft/microsoft-ui-xaml/issues/6521

hub/apps/winui/winui2/getting-started.md
```diff
@@ -54,30 +54,28 @@ The library is available as a NuGet package that can be added to any new or exis
...
+                </controls:XamlControlsResources.MergedDictionaries>
+            </controls:XamlControlsResources>
         </Application.Resources>
 
     </Application>
     ```
 
-    > [!IMPORTANT]
-    > The order of resources added to a ResourceDictionary affects the order in which they are applied. The `XamlControlsResources` dictionary overrides many default resource keys and should therefore be added to `Application.Resources` first so that it doesn't override any other custom styles or resources in your app. For more information on resource loading, see [ResourceDictionary and XAML resource references](/windows/uwp/design/controls-and-patterns/resourcedictionary-and-xaml-resource-references).
-
```
marcelwgn, line 80: Why remove tis section, it's not wrong and can help people not run into some of the pitfalls with the XamlControlsResources dictionary.

sylveon: I removed it since it doesn't apply to the new guidance (there is no MergedDictionaries ordering here). Maybe we should move it somewhere else that would be more appropriate? The linked article sounds like a good place to put it in.

marcelwgn: Fair point, I guess it makes sense to put an updated version in the linked article and remove it from this document with the current state of the PR.

ktoliver:
Hi @Karl-Bridge-Microsoft - Could you take a look at the proposed update? Thanks!

Karl-Bridge-Microsoft (approved):

Karl-Bridge-Microsoft:
I'll look into moving the detail in the note to the suggested topic in a separate PR.

bogdan-patraucean:
Hello! I tried with this documentation to do a style extension in my app but I'm getting an error using this code (same as in documentation):

```
<controls:XamlControlsResources>
            <controls:XamlControlsResources.MergedDictionaries>
                <ResourceDictionary Source="/Resources/Resources.xaml"/>
                <ResourceDictionary Source="/Resources/Styles.xaml"/>
            </controls:XamlControlsResources.MergedDictionaries>
            <vm:ViewModelLocator x:Key="VML"></vm:ViewModelLocator>
</controls:XamlControlsResources>
```
The error is: **Cannot find a Resource with the Name/Key DefaultListViewItemStyle**

I'm trying to create a new style based on DefaultListViewItemStyle in my Styles.xaml file, like this: 

`<Style x:Key="ResetListViewItemStyle" TargetType="ListViewItem" BasedOn="{StaticResource DefaultListViewItemStyle}"> `

Should I open a new issue with this? or am I doing something wrong? maybe this is not applying to WinUI 3? as I'm using WindowsAppSDK 1.0.

PR #3740, opened 2022-03-23
Remove DisplayRequest from "not supported" for WinAppSDK

angelazhangmsft:
DisplayRequest had a bug in Windows 11, but it was fixed. See https://github.com/microsoft/CsWinRT/issues/962

It is by default supported on desktop apps, so doesn't need to be called out here.


PRMerger18:
@angelazhangmsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

stevewhims (approved):
Thank you!

PR #3752, opened 2022-03-28
Update installer link from WinUI3 tutorial

angelazhangmsft:
Instead of pointing to a specific version of the 1.0 installer, point to the downloads page instead.

Developers may want to download either the latest 1.0 installer or 1.1 installer in the future, so it doesn't make sense to link directly to the file download (and this avoids duplication of download links)

Merge anytime


PRMerger10:
@angelazhangmsft : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Karl-Bridge-Microsoft (approved):

PR #3760, opened 2022-03-31
Fix sub commands in "update" section

Kade-N:
The "update all" section mistakenly left out that "update" is a sub-command of "source", as it does not work on its own as a top level winget command.

Additionally, these two update commands listed "C:\" at the beginning, which naturally only works if winget is located at the root of the C drive, which is not normally the case, nor do the other examples list this - thus, this has been removed.

### DIFF
--- a/hub/package-manager/winget/source.md
+++ b/hub/package-manager/winget/source.md
@@ -113,11 +113,11 @@ usage: `winget source update [-n, --name] \<name>`
 
 ### update all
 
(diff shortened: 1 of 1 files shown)


PRMerger7:
@Kade-N : Thanks for your contribution! The author(s) have been notified to review your proposed change.

mattwojo:
Thank you @Kade-N! Merging!

PR #3938, opened 2022-07-12
Update BoxPanel custom panel example

michael-hawker:
Used this as an example for a template we're building in the Windows Community Toolkit and ran into some issues with the sample and code. Figured I'd contribute back to the article here the fixes I had to do.

Fixes:
- Off-by-one logic error for cutting size of box (for instance try '6' as a value, should be 2x3 not 3x3 still) (from my understanding of the scenario)
- Missing `{` in `LimitUnboundedSize` method
- Missing `;` in `ArrangeOverride` method
- Improper definition of a Dependency Property (guidelines are strict here about property and DP names aligning otherwise issues can occur, picked a more standard XAML based Panel name here as well for this scenario.
  - Also added changed callback to invalidate layout so that the dependency property change actually updates the layout of the control immediately.
- Provided more context for where the adjustment to the code goes.

### DIFF
--- a/hub/apps/design/layout/boxpanel-example-custom-panel.md
+++ b/hub/apps/design/layout/boxpanel-example-custom-panel.md
@@ -95,11 +95,11 @@ protected override Size MeasureOverride(Size availableSize)
     if (aspectratio > 1)
     {
         rowcount = maxrc;
-        colcount = (maxrc > 2 && Children.Count < maxrc * (maxrc - 1)) ? maxrc - 1 : maxrc;
+        colcount = (maxrc > 2 && Children.Count <= maxrc * (maxrc - 1)) ? maxrc - 1 : maxrc;
     } 
     else 
     {
-        rowcount = (maxrc > 2 && Children.Count < maxrc * (maxrc - 1)) ? maxrc - 1 : maxrc;
+        rowcount = (maxrc > 2 && Children.Count <= maxrc * (maxrc - 1)) ? maxrc - 1 : maxrc;
         colcount = maxrc;
     }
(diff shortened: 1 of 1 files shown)


PRMerger8:
@michael-hawker : Thanks for your contribution! The author(s) have been notified to review your proposed change.

Jak-MS:
@jwmsft 

Can you review this PR?

IMPORTANT: When this content is ready to merge, you must add `#sign-off` in a comment or the approval may get overlooked.


#label:"aq-pr-triaged"
@MicrosoftDocs/public-repo-pr-review-team

jwmsft (approved):

jwmsft:
@michael-hawker, thank you!

@sign-off

michael-hawker:
Thanks folks! Realized after there are some other optimizations around not using so many local variables in the example, but can do that as a separate thing later, if desired.

I know it's also probably to reduce the complexity of the code a bit, but at the same time I feel like it's good to promote best practices. Let me know if you'd like me to open another PR for that after.

jwmsft:
> Thanks folks! Realized after there are some other optimizations around not using so many local variables in the example, but can do that as a separate thing later, if desired.
> 
> I know it's also probably to reduce the complexity of the code a bit, but at the same time I feel like it's good to promote best practices. Let me know if you'd like me to open another PR for that after.

@michael-hawker, thanks for this. It's always a balance between keeping code simple for the docs and showing best practices. For a complete, complex example like this, I think it's okay to make it more complex and show the best practices. Please feel free to open another PR with your optimizations.

PR #4018, opened 2022-08-26
Update data-binding-quickstart.md

ThomasLeopold2022:
Documentation text referred to a probably old property name.


alvinashcraft (approved):
Thanks for catching this docs issue!

PR #4052, opened 2022-09-17
Update text-extractor.md

crutkas:


mattwojo (approved):
Made a small update and added a support link to language settings. Otherwise looks good.

PR #4056, opened 2022-09-21
Update sqlite-databases.md

Fincap:
Removed redundant db.Close() calls in AddData and GetData C# examples.
SqliteConnection class implements IDisposable and executes SqliteConnection.Close() as part of its IDisposable.Dispose() call, example contains redundant line as instance is disposed implicitly at the end of the using block.
InitializeDatabase example on line 175 already demonstrates this.


alvinashcraft (approved):

alvinashcraft:
Looks good. Thanks for this update!

PR #4099, opened 2022-10-18
Minor edits to reflect actual values

philnach:
Minor edits to correct some inaccuracies on the page.

Android Studio typically generates an 'app-debug.apk' filename, removed the space Other updated properties are taken from the current preview releases of Windows Subsystem for Android


mattwojo:
Thank you @philnach! Merging. :)

PR #4100, opened 2022-10-19
Update install.md

Spyderdreams:
winget-install was missing dash


mattwojo:
Good catch! Thank you @Spyderdreams!

mattwojo:
Oops - just looked at this again and this is not actually a correct update. No dash in the winget command. (Reverting)

PR #4106, opened 2022-10-26
fix: nit: JavaScript case

SphinxKnight:


SphinxKnight:
After cloning locally, I found some other occurrences, I'll open another PR to tackle them all.
Sorry for the noise.

PR #4138, opened 2022-11-10
Updating docs to remove preview labels for .NET and VS

jamshedd:


mattwojo (approved):
Looks great! Thanks @jamshedd!

PR #4182, opened 2022-12-07
Pcd 3953 rss reader tutorial

### LINKED ISSUE #3953: Update Rust examples to use the latest "windows" crate
The current revision of [RSS reader tutorial (Rust for Windows with VS Code)](https://github.com/MicrosoftDocs/windows-uwp/blob/7d5eda5ccd6645de974b101a345fbb12bf4c0beb/hub/dev-environment/rust/rss-reader-rust-for-windows.md) uses a very old version of [windows-rs](https://github.com/microsoft/windows-rs/).

While this is totally fine, newer versions of the crate uses Cargo features and overall looks cleaner because it precludes build.rs

Cargo.toml snippet with the latest crate for the RSS example:
```toml
[dependencies.windows]
version = "0.39"
features = [
    "Foundation_Collections",
    "Web_Syndication",
]
```
And the RSS code which remains mostly the same besides the imports and the w! macro:
```rust
use windows::{
    core::{w, Result},
    Foundation::Uri,
    Web::Syndication::SyndicationClient,
};

fn main() -> Result<()> {
    let uri = Uri::CreateUri(w!("https://blogs.windows.com/feed"))?;
    let client = SyndicationClient::new()?;
    let feed = client.RetrieveFeedAsync(&uri)?.get()?;

    for item in feed.Items()? {
        println!("{}", item.Title()?.Text()?);
    }

    Ok(())
}
```
And the MessageBoxA example:
```rust
use windows::{
    s,
    Win32::UI::WindowsAndMessaging::{MessageBoxA, MB_OK},
};

fn main() {
    unsafe {
        MessageBoxA(None, s!("Text"), s!("Caption"), MB_OK);
    }
}
```

Paul-Dempsey:
Fixes #3953. Notes:

- Removes an unused image (I grepped everything under hub to see if there were any references).

- Both screen snips used in the page are updated.

- I've tried to make the examples as close as possible to the official samples in the windows-rs public repo as of 0.43.0. The message_box sample at the **tip** of windows-rs does not compile with 0.43.0, due to updates in the APIs, so I've omitted the problematic call to ShellMessageBoxW. The example in this update works with any recent version. When the next windows crate is published, the sample could be updated to reflect the latest sample in the windows-rs repo.


Paul-Dempsey:

stevewhims (approved):

PR #4193, opened 2022-12-12
Fix bad doc guidance around auth

bgavrilMS:
To refresh an app token, you simply make the same call to the AAD endpoint again. The "refresh" flow you recommend there is not usable in this context because you have to auth code.

Source: I am the EM of the Identity SDK for .NET


hickeys (approved):

hickeys:
#sign-off

PR #4205, opened 2022-12-18
Update rss-reader-rust-for-windows.md

jaymichaud:
Add missing ampersand before "uri" argument in step 8 code. This omission makes the sample fail to compile. It is correct in step 9, but for students following along and adding code step-by-step, this is easy to miss.


stevewhims (approved):
Thank you!

PR #4213, opened 2022-12-31
Update "MS Edge" in Apps for Websites to "MS Edge Legacy"

dongle-the-gadget:
This capability is only supported in Edge Legacy (EdgeHTML), not Edge Chromium.


alvinashcraft (approved):
Thanks for updating this note!

Issue #4164, opened 2022-11-25
Time and date direct activation should be changed from `)` to `(`


Eragra3:
Time and date plugin shows `)` (closing parenthesis) character as direct activation command.
On my machine it seems to be `(` (opening parenthesis).

Maybe it got changed in newer version of PowerToys?

---
#### Document Details

⚠ *Do not edit this section. It is required for learn.microsoft.com ➟ GitHub issue linking.*

* ID: 0b951899-69ca-9057-f799-14491e1b0a74
* Version Independent ID: 18e0202f-3e91-a5b9-62f2-6cd40e5661a8
* Content: [PowerToys Run utility for Windows](https://learn.microsoft.com/en-us/windows/powertoys/run)
* Content Source: [hub/powertoys/run.md](https://github.com/MicrosoftDocs/windows-dev-docs/blob/docs/hub/powertoys/run.md)
* Product: **dev-environment**
* Technology: **powertoys**
* GitHub Login: @mattwojo
* Microsoft Alias: **mattwoj**

Jay-o-Way:
`(` was changed because it had too many conflicts. @Eragra3 what version are you using?

@mattwojo In other words: **Docs are correct.**
![image](https://user-images.githubusercontent.com/65828559/204415652-a50edfb8-d003-40cd-bff3-b47743cb80a7.png)

Jay-o-Way:
See

- #3957
- https://github.com/microsoft/PowerToys/issues/18490
- https://github.com/microsoft/PowerToys/pull/19593

We've included a message at https://learn.microsoft.com/windows/powertoys/run#direct-activation-commands for this exact situation.

Eragra3:
@Jay-o-Way Hi, sorry I didn't respond earlier.
That's interesting, because it's still `(` for me. I am using latest version
![obraz](https://user-images.githubusercontent.com/9082983/213536600-a865c86c-bdec-4bf1-8f7d-a2a66ee03cc2.png)
And this is what I get
![obraz](https://user-images.githubusercontent.com/9082983/213536967-6c47b42e-8e7e-43a7-b67e-ee142ca0dec7.png)


I have mixed region settings, you can see it shows results in polish, I don't know if that's why.
Also I have used PowerToys for quite long time, so maybe something is lingering from previous versions.

Jay-o-Way:
@Eragra3 probably: the setting was kept from the older settings for the app. Actively changing existing "keywords" was probably too disruptive.

[...]

PR #3957, opened 2022-07-23
[PowerToys] Docs update for activation command change

htcfreek:
Docs update for changes in PR https://github.com/microsoft/PowerToys/pull/19593 (issue https://github.com/microsoft/PowerToys/issues/18490).
- Updating keyword for `Time and Date`
- Adding a hint about conflicting keywords

image

**Note: Please wait with merging until the approval from @jaimecbernardo or @crutkas !!**

PR #4231, opened 2023-01-11
Update run.md

luis-aranda:
The documentation in the examples for the time plugin is using the ")" character for the activation command which is incorrect. The correct character is "("

### DIFF
--- a/hub/powertoys/run.md
+++ b/hub/powertoys/run.md
@@ -227,10 +227,10 @@ The Time and date plugin provides the current time and date or a custom one in d
 
 Examples:
 
-- `time` or `) time` to show the time.
-- `) 3/27/2022` to show all available formats for a date value.
-- `) calendar week::3/27/2022` to show the calendar week for a date value.
-- `) unix epoch::3/27/2022 10:30:45 AM` to convert the given time and date value into a Unix epoch timestamp.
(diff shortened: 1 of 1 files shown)


Jay-o-Way:
@luis-aranda what is the version of PowerToys you are using? It might be heavily outdated.
@mattwojo the correct char is `)` because (in the past it was) `(` had many conficts. Ergo, **this pr is invalid**. See also #4164

luis-aranda:
@Jay-o-Way  I am using v0.66.0

image

If you see the activation char for me is (

image

) does nothing on my run prompt

image

Jay-o-Way:
@luis-aranda okay, maybe that's a left-over from the past. The _default_ (for new installations) is changed.
See

- #3957
- https://github.com/microsoft/PowerToys/issues/18490
- https://github.com/microsoft/PowerToys/pull/19593

We've included a message at https://learn.microsoft.com/windows/powertoys/run#direct-activation-commands for this exact situation.

Jay-o-Way:
@mattwojo would you like to close this pr please?

mattwojo:
Hmm... we may want to add a note about when this command changed in a note. I'll do some investigating. Thanks for filing @luis-aranda. Closing this as it looks like the info for the current version of the tool represented is correct in the doc.

PR #4234, opened 2023-01-17
Fix typo pointing to BC1 instead of BC4

fdwr:
BC4 is the format which uses 1-byte per each of the two endpoints along with 16 x 3-bit interpolation indices, not BC1 (which uses 5:6:5).


stevewhims (approved):

stevewhims:
The errors above are not directly related to the change (which is good), so I'll complete this PR. If there are subsequent issues with this change making its way to the live site, then I'll take care of them.

Thanks! :)
-Steve

PR #4235, opened 2023-01-19
fix age and xboxTitleId attributes for groupby

light-le:
The original `ageGroup` and `xboxTitleIdHex` did not work with the api. The correct ones should be `age` and `xboxTitleId`


hickeys (approved):

hickeys:
#sign-off

PR #369, opened 2018-03-21
Fix download sample links

laedit:

PR #4237, opened 2023-01-19
Clarify rules on third-party product mentions

rfreytag:
Microsoft App Store policy currently bans product names mentioning third-party products [1]. Currently this rule is inconsistently applied by Microsoft Store app reviewers because it isn't clear.

Non-deceptive usage of third-party brands with products is fair use by Trademark rules [2]. Furthermore, products supporting third-party products must emphasize their supporting/dependent role as shown by many products in the Microsoft Store that incorporate third party products in their name [3].

This policy is not consistently applied by the Microsoft Store app reviewers. Many apps use third-party products in their names; yet two many-years-old products of ours recently were rejected because of this rule. Clearly, either the rule must change or all products breaking this Store policy seem to require a name change (a bad choice IMO).

Please change Store policy to italicize only the third-party product name in Microsoft Store product titles to clarify they are not part of the actual name. Also include a disclaimer immediately in the product description declaring “&lt;third-party product&gt; is not a product of &lt;publisher name&gt;”.

References:

1. See second bullet under 10.1 “Your product must not use a name, images, or any other metadata that is the same as that of other products unless the product is also published by you.” (https://learn.microsoft.com/en-us/windows/apps/publish/store-policies#101-distinct-function--value-accurate-representation)

2. Fair use permits use of another’s trademark to refer to the trademark owner’s goods and services associated with the mark.” (https://www.inta.org/fact-sheets/fair-use-of-trademarks-intended-for-a-non-legal-audience/)

3. Here are products currently in the Microsoft Store that accurately and fairly use the name of a third-party product (names are in the URLs)…<please see #3 here: https://stackoverflow.com/questions/75148703/third-party-product-names-banned-in-microsoft-store-products-names-but-customers> because this forum limits the number of URLs in a posting.

### DIFF
--- a/hub/apps/publish/store-policies.md
+++ b/hub/apps/publish/store-policies.md
@@ -81,7 +81,7 @@ If you have feedback on the policies, please let us know by commenting in [our f
 
 - Your product title or name must be unique and must not contain marketing or descriptive text, including extraneous use of keywords.
 
-- Your product must not use a name, images, or any other metadata that is the same as that of other products unless the product is also published by you.
+- Your product description must clearly cite the ownder of any trademark or product that is not owned by you.  
 
 - The value proposition of your product must be clear during the first run experience.
 
--- a/hub/apps/publish/store-policies.md
+++ b/hub/apps/publish/store-policies.md
@@ -81,7 +81,7 @@ If you have feedback on the policies, please let us know by commenting in [our f
 
 - Your product title or name must be unique and must not contain marketing or descriptive text, including extraneous use of keywords.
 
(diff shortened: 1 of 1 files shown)


rfreytag:
@hickeys - I have no further edits to this pull request (#4237) or the associated issue (#369).

hickeys:
Thank you for your feedback, but I cannot approve this PR. As it says in the document, the proper place for feedback is [this forum](https://learn.microsoft.com/en-us/answers/topics/partner-center-general.html).

PR #4238, opened 2023-01-19
Update docs based on feedback from customers and feature PM

DaveVoyles:
After speaking with feature PM Johnathan Garrigues, I have more clarification on how this works. Gradual rollout affects only which set of packages are installed.  It does not affect who can see the PDP (Product Display Page in the store).

Partner Center had questions coming from customers who were unclear as to whether package rollout applies to existing customers or new customers.

To make it clear I have replaced "App's customers" with "customers who have your package installed" where necessary.


hickeys (approved):

hickeys:
#sign-off

PR #4248, opened 2023-01-23
Missing paragraph on MFC

rtischer8277:
Let me explain: I am developing a large framework. I started out developing it using MFC and I will be finishing it using MFC. I have experimented with all kinds of GUI and other libraries in between, but have ultimately discarded them basically because the layers of indirection at some point caused existential app development problems. Two factors have emerged that have kept me programming in MFC (and by the way kept my app from failing): long-term code maintenance and stability, and the unfettered access-to-code capability. By that I mean continued access to underlying Windows OS functions, not to mention C++'s native ability to abstract and encapsulate without indirection costs including loss of code ownership.

To not mention MFC in this article would be a serious oversight in my opinion, despite its reputation as having no future.

By the way, in my suggested paragraph text, I use the IPv4/IPv6 example to drive home my point about MFC being able to continually get out of tight programming spots by utilizing the powerful underlying forest of OS function calls. Also, I can back up my statements with code since I did just that. I took 2 years to develop an underlying IPv6 inter-connectivity capability from within an MFC app. This is just one real-world example of how MFC can't get stuck.


rtischer8277:
I am unable to obtain access to the documentation that tells me how to make a relative URL path. Neither of my two Microsoft accounts works for which I have user names and passwords. They are my Microsoft account and Github account. Any suggestions?

rtischer8277 (commented):
Is 'Missing paragraph on MFC #4248' destined for the stale bin? Or is it still being considered?

ShannonLeavitt:
@QuinnRadich - Could you review this PR?

IMPORTANT: When the changes are ready for publication, add a `#sign-off` comment to signal that the PR is ready for the review team to merge. 

#label:"aq-pr-triaged"
@MicrosoftDocs/public-repo-pr-review-team

QuinnRadich:
MFC falls outside of the scope of docs in this repo. Documentation on MFC exists elsewhere in our doc ecosystem, but it's an older framework that won't be detailed here. Devs looking for information on it know how to find it, but we won't be promoting it on this page.

If you have particular feedback on the MFC docs, you can file it over where they live: https://learn.microsoft.com/en-us/cpp/mfc/mfc-desktop-applications?view=msvc-170

rtischer8277:
Obviously, I don't agree, but I do appreciate getting an answer. Thanks.

PR #4251, opened 2023-01-24
Update plugin time and date documentation

jcaldeira:
Update direct activation command on plugin time and date to match actual default when PowerToys it's installed. `)` &rarr; `(`


mattwojo:
LGTM. @crutkas - Ok to merge?

crutkas:
https://github.com/microsoft/PowerToys/blob/main/src/modules/launcher/Plugins/Microsoft.PowerToys.Run.Plugin.TimeDate/plugin.json

Actual default is different now. When we first launched, it was ( but it conflicted with math related queries. 

Existing doc is correct

mattwojo:
Gotcha. Thanks for filing @jcaldeira. Sorry for the confusion -- sounds like the current command will work in the most recent version. Closing.

PR #4255, opened 2023-01-26
Update instructions for converting a project to use Single-project MSIX Packaging

evelynwu-msft:
It is no longer necessary to set the `AppContainerApplication` property to `true` when converting a C++ WinAppSDK app using Windows Application Packaging to Single-project MSIX Packaging and will in fact lead to incorrect behavior. This PR updates the documentation accordingly.


stevewhims (approved):

stevewhims:
Yes, I tested the tutorial with this change and all is well. Thank you! :)

PR #4272, opened 2023-01-31
winget is not pre-installed on Win11

### LINKED ISSUE #4212: winget command-line tool is not bundled in modern versions of Windows 10 contrary to what documentation says
https://github.com/MicrosoftDocs/windows-dev-docs/blob/docs/hub/package-manager/winget/index.md#install-winget says

> Windows Package Manager winget command-line tool is bundled with Windows 11 and modern versions of Windows 10 by default as the App Installer.

I have updated to latest version of Windows 10: version 22H2 (OS Build 19045.2364).

But, typing `winget` in PowerShell says the term is not recognized.

zivkan:
Numerous times I've created Virtual Machines in Azure, and on at least two different physical machines with Windows 11 pre-installed, `winget` was not pre-installed. I had to install the App Installer from the Store app in every instance. Therefore, I believe that claiming winget is bundled with Windows is factually incorrect.

fixes: https://github.com/MicrosoftDocs/windows-dev-docs/issues/4212

### DIFF
--- a/hub/package-manager/winget/index.md
+++ b/hub/package-manager/winget/index.md
@@ -12,9 +12,9 @@ The **winget** command line tool enables users to discover, install, upgrade, re
 
 ## Install winget
 
-Windows Package Manager **winget** command-line tool is bundled with Windows 11 and modern versions of Windows 10 by default as the **App Installer**.
+Windows Package Manager **winget** command-line tool is available on Windows 11 and modern versions of Windows 10 as the **App Installer**.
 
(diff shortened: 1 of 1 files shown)


mattwojo:
Thanks @zivkan. I like the way you reworded this. 
Winget is budled with the App Installer, the issue is that it is only *registered* as part of an asynch process after the user logs in for the first time. I'll add a note about this as well.
There is a 'trick' to get it registered faster by running this PowerShell script:
`Add-AppxPackage -RegisterByFamilyName -MainPackage Microsoft.DesktopAppInstaller_8wekyb3d8bbwe`

PR #4298, opened 2023-02-09
Update index.md

alexsolaat:
Added "." to line 40, as invoking file would not work as-provided unless refactored or renamed file


mattwojo:
Thank you for this contribution @alexsolaat. This issue was fixed in https://github.com/MicrosoftDocs/windows-dev-docs/commit/ca90ce0471458f85af6dc789200b263affcfd8ec

PR #4318, opened 2023-03-01
Update install.md

crutkas:
removing unneeded install requirements


Jay-o-Way:
We don't neeed to mention what's packaged "inside" the installer, right?

mattwojo:
Just let me know once this is good to go and I'll merge it in.

PR #4320, opened 2023-03-03
Update .Net version

JeroenOortwijn:
Commit ed1000ee90f8a6b8343dbc09b774b392176fa9dc updated the .NET versions throughout this document, but forgot to change this one.

### DIFF
--- a/hub/apps/desktop/modernize/desktop-to-uwp-enhance.md
+++ b/hub/apps/desktop/modernize/desktop-to-uwp-enhance.md
@@ -53,7 +53,7 @@ This option is supported only in projects that use .NET 6 (or later) and target 
     <TargetFramework>net6.0-windows10.0.19041.0</TargetFramework>
     ```
     
-    In later versions of .NET, you can replace the value with the relevant version, for example **net6.0-windows10.0.19041.0**.
+    In later versions of .NET, you can replace the value with the relevant version, for example **net7.0-windows10.0.19041.0**.
 
 3. Save your changes and close the project file.
 


stevewhims:
Thanks, @JeroenOortwijn! The commit you mention was changing "net5" to "net6", because .NET 5 was out of support. But the change you're making here ("net6" to "net7") is a new idea; it wasn't forgotten in that commit.

I think we can stick with "net6" for a good while yet. So I'll go ahead and close this PR and we can revisit the idea in the future.

Thanks!
-Steve

JeroenOortwijn:
I think I wasn't clear enough.
Before commit ed1000ee90f8a6b8343dbc09b774b392176fa9dc, the text was like:
```
For example, the following element is for a project that targets Windows 10, version 2004.
<TargetFramework>net5.0-windows10.0.19041.0</TargetFramework>
In later versions of .NET, you can replace the value with the relevant version, for example net6.0-windows10.0.19041.0.
```
As you can see, in the example it talks about net5.0 and in the last sentence it talks about net6.0.

After commit ed1000ee90f8a6b8343dbc09b774b392176fa9dc, the text is now:
```
For example, the following element is for a project that targets Windows 10, version 2004.
<TargetFramework>net6.0-windows10.0.19041.0</TargetFramework>
In later versions of .NET, you can replace the value with the relevant version, for example net6.0-windows10.0.19041.0.
```
As you can see, they now both talk about net6.0, which kinda defeats the purpose of the last sentence.

My PR changes the last sentence to net7.0, so it makes sense again. (Although the wording of this sentence is kinda strange and could be improved I think, but I'm not a native speaker.)

PR #4327, opened 2023-03-08
A couple of spelling/reference corrections

nocusds:
A couple of spelling/reference corrections


alvinashcraft (approved):
Thanks for catching these issues! Looks good.

PR #4351, opened 2023-03-22
Fix typo in resume-an-app.md

krschau:
"may be been" -> "may have been"


alvinashcraft (approved):
Good catch. Thank you!

PR #4352, opened 2023-03-22
Update suspend-an-app.md

krschau:
Sentence seems like it's missing some words. "A deferral doesn't increase the amount [of what?] you have run..."


alvinashcraft (approved):
Thank you. This change makes it make more sense.

PR #4381, opened 2023-04-12
Update index to make wording more natural

eelnivla:


ShannonLeavitt:
Looks good. Thanks @eelnivla

PR #4520, opened 2023-07-07
Update WinGet supported commands

mdanish-kh:


mattwojo:
Thank you for adding these @mdanish-kh! Merged.

PR #4526, opened 2023-07-12
Update arguments and options of WinGet commands

mdanish-kh:
Added new arguments / options that are supported on the latest 1.5 stable release of WinGet.

Here is a summary of the changes:

- Changed order of commands to match as they appear in the WinGet client to make this easy to review if you see the corresponding WinGet client outputs
- Added new options / modified existing ones to be up to date with the WinGet client
- `-?, --help` is classified as an option not an argument in most commands, so I have changed that accordingly in the docs
- Trimmed trailing whitespaces from the markdown files
- Add arguments and options for each subcommand of `winget source` to be consistent with how they are listed in other commands, say `winget pin`

### DIFF
--- a/hub/package-manager/winget/export.md
+++ b/hub/package-manager/winget/export.md
@@ -25,7 +25,7 @@ The **export** command is often used to create a file that you can share with ot
 The following arguments are available.
 
 | Argument    | Description |
-|-------------|-------------|  
+|-------------|-------------|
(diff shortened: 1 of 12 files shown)


denelon (approved):

mattwojo (approved):

mattwojo:
Looks good -- thank you for making these contributions to improve the docs @mdanish-kh!

PR #4548, opened 2023-07-30
Add aliases for WinGet Commands

mdanish-kh:
To verify the added aliases, you can do `winget <command> --help` that'll show all the aliases available for that particular command.


mattwojo:
Looks good. Thank you for adding these aliases!

PR #4550, opened 2023-07-30
Add section for exit codes in troubleshooting

mdanish-kh:


mattwojo:
Thank you for adding this link! Good call to include it in the docs.

PR #4560, opened 2023-08-04
Update create.md

JohnMcPMS:
Remove false statement:
```
Each resource must include **directives** and **settings**.
```

Remove following statement that seemed awkward and not needed without previous statement.
```
Optionally, it can also include an **id** value.
```


denelon (approved):

mattwojo:
Thanks for this update @JohnMcPMS!
Merged!

PR #4582, opened 2023-08-21
WSA redirect to Windows apps

kevinkieselbach:
Added a "redirect to Windows apps" section to the WSA page describing how to launch Windows file/protocol handlers from Android apps. This includes WSA's implicit redirection and the new LAUNCH_URI intent action.

I also removed a description of WSA's software-backed file encryption since we removed that support due to userdata corruption issues and lack of real security without a hardware keystore.

I also added (TM) superscripts to all name drops of WSA in the documentation since we were doing that inconsistently. We don't use the (TM) superscript on any name drop of Android outside of the WSA name, so we should probably do that, but it wasn't clear if that was intentional since we consistently didn't do it.

### DIFF
--- a/hub/android/wsa/index.md
+++ b/hub/android/wsa/index.md
@@ -1,6 +1,6 @@
 ---
 title: Windows Subsystem for Android™️
-description: Windows Subsystem for Android enables your Windows 11 device to run Android applications that are available in the Amazon Appstore. Learn more about set up, device requirements, installing and uninstalling, input compatibility, testing and debugging, and using the Amazon Appstore.
+description: Windows Subsystem for Android™️ enables your Windows 11 device to run Android applications that are available in the Amazon Appstore. Learn more about set up, device requirements, installing and uninstalling, input compatibility, testing and debugging, and using the Amazon Appstore.
 author: mattwojo
 ms.author: mattwoj
 ms.reviewer: mousma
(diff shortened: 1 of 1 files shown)


hub/android/wsa/index.md
```diff
@@ -226,21 +226,36 @@ The properties for the Windows Subsystem for Android VM are listed below. Hardco
...
 
-Windows Subsystem for Android performs per-file encryption that is software-based.
+```java
+Intent intent = new Intent("com.microsoft.windows.LAUNCH_URI");
+intent.putExtra("com.microsoft.windows.EXTRA_URI", "ms-calculator:");
+ 
+try {
+    startActivity(intent);
+} catch (ActivityNotFoundException e) {
+    // Not running in Windows Subsystem for Android™️ (or running on an old build).
```
jaholme: Be a bit more verbose "(or running on an older build that did not contain this feature)"

hub/android/wsa/index.md
```diff
@@ -226,21 +226,36 @@ The properties for the Windows Subsystem for Android VM are listed below. Hardco
...
+}
+```
+
+## Security
 
 Both Windows kernel-mode drivers and Windows applications running at medium integrity level (IL) can inspect arbitrary Android containers and Android app memory. There are no plans to add detection for cheats/macro/bot/suspicious behaviors detection in the short-term.
 
 Developers querying `getSecurityLevel` will get `SECURITY_LEVEL_SW_SECURE_CRYPTO`. Learn more about `getSecurityLevel` in the [Android API Reference guide](https://developer.android.com/reference/android/media/MediaDrm#getSecurityLevel(byte[])).
```
jaholme, line 250: Is this still true?

jaholme (approved):

hub/android/wsa/index.md
kevinkieselbach: I think so.  This is a MediaDrm API, so I think it's referring to things like WideVine.  I don't think File Based Encryption is related to DRM.

jaholme: Makes sense I didn't see the mediadrm portion of the URL.

mattwojo:
Thanks for making these updates @kevinkieselbach! Merging them in - should be live within an hour.

PR #4607, opened 2023-09-08
Update display-ui-objects.md

Fulgen301:
`IDataTransferManagerInterop` is not related to `UserConsentVerifier` and seems to be a copy-paste error from the paragraph above it. `IUserConsentVerifierInterop` is the correct interface.


stevewhims (approved):
Thank you!

PR #4628, opened 2023-09-24
Update index.md

PesBandi:
Changed "Processor support" to "Processor architecture support". To me, "Processor support" sounds like a list of all supported processor and since this only mentions x64 and ARM64, I think that "Processor architecture support" is more appropriate.


alvinashcraft (approved):
This is a good change. Thank you!

PR #4637, opened 2023-09-27
Update index.md

PesBandi:
Since the utility is called "File explorer add-ons", it doesn't make sense that only "File explorer" is a link.


alvinashcraft (approved):
This change makes sense to me. Thank you!

PR #4643, opened 2023-09-29
Update interop-winrt-abi.md

justanotheranonymoususer:
Code examples above use static_cast and it's probably safer. Maybe other places can be changed too.

### DIFF
--- a/uwp/cpp-and-winrt-apis/interop-winrt-abi.md
+++ b/uwp/cpp-and-winrt-apis/interop-winrt-abi.md
@@ -304,8 +304,8 @@ static_assert(std::is_same_v<winrt::default_interface<winrt::Sample>, winrt::ISa
 
 | Operation | How to do it | Notes |
 |-|-|-|
-| Extract **ISample\*** from **winrt::Sample** | `p = reinterpret_cast<ISample*>(get_abi(s));` | *s* still owns the object. |
-| Detach **ISample\*** from **winrt::Sample** | `p = reinterpret_cast<ISample*>(detach_abi(s));` | *s* no longer owns the object. |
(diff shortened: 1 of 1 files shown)


stevewhims:
Thanks, I'm investigating. This section is about unsafe interop, so I'm not yet sure we want to take this change.

justanotheranonymoususer:
So you prefer to keep it as unsafe as possible? =)

justanotheranonymoususer:
I can also submit a change which makes the other places less safe. At least it will be consistent

oldnewthing:
The operation is inherently unsafe regardless of which cast you use. If you cast to the wrong thing, `safe_cast` will still succeed (since `void*` can be safe-casted to anything). Using `safe_cast` is misleading because it makes the operation look safer than it is. `reinterpret_cast` more accurately reflects how dangerous the operation is.

justanotheranonymoususer:
What is safe_cast and where did you see it?
Anyway, see https://stackoverflow.com/a/310489
And also again, now the doc is a mix of both

oldnewthing:
Sorry, I meant static-cast, not safe-cast. But the underlying principle applies. These casts are not safe. They should look scary.

I would be in favor of changing the existing static_casts to reinterpret_cast. They are all dangerous.

stevewhims:
Thanks, @justanotheranonymoususer, for getting me thinking about this aspect of the topic.

What @oldnewthing says resonates with me. That either syntax is unsafe; but that choosing `reinterpret_cast` has the benefit of telegraphing the inherent unsafeness. So I've left the `reinterpret_cast`s as-is, and (in a separate PR in our private repo), I've changed the remaining 3 `static_cast`s to `reinterpret_cast`. That change is live now. I'll go ahead and close this one.

Thanks again!
-Steve

PR #4665, opened 2023-10-11
Update GPO docs for PowerToys

htcfreek:
This PR replaces the list of policies with an link to the docs on GitHub. I don't see any reason why we should maintain two lists.

**Closes:** https://github.com/microsoft/PowerToys/issues/29151

@crutkas, @jaimecbernardo 
I used an aka link. Hope this is correct doing.
Please create the following link: `aka.ms/powertoys-gpo-list` => `https://github.com/microsoft/PowerToys/blob/stable/doc/gpo/README.md#policies`

### DIFF
--- a/hub/powertoys/grouppolicy.md
+++ b/hub/powertoys/grouppolicy.md
@@ -1,7 +1,7 @@
 ---
 title: PowerToys Group Policy
 description: Group policy objects 
-ms.date: 08/03/2023
+ms.date: 10/11/2023
 ms.topic: article
 no-loc: [PowerToys, Windows, Group Policy, Win]
 ---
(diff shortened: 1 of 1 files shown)


crutkas:
I'd then reroute https://aka.ms/PowerToysOverview_GPO to the learn site and do "Docs for GPO have moved to Learn.microsoft ....

crutkas:
https://github.com/microsoft/PowerToys/pull/29190 would be the docs i would move over with the smidge of adjustemnt

crutkas:
@alvinashcraft what are your thoughts here.  which approach do you think is better.

alvinashcraft:
> @alvinashcraft Alvin Ashcraft FTE what are your thoughts here. which approach do you think is better.

I agree that it makes sense to move the GPO Policies information to the docs on the Learn site and reference it from the readme file. It feels like it currently has more info than you'd typically put in a readme, and it should reach a wider audience if it's moved to the Learn docs.

htcfreek:
@crutkas , @alvinashcraft 
Then let's move it completely. Two place with information makes no sense.

alvinashcraft:
[like]  Alvin Ashcraft reacted to your message:
________________________________

htcfreek:
Closed the PR as I can't work on it the next weeks. @crutkas please create a new one after you have updated the readme in our GH repository.

PR #4678, opened 2023-10-14
Improve Windows Dev Options page

ahmed605:
Changes:
- Corrected the **.NET Runtime** row for UWP.
- Corrected the **Fluent Design** row for UWP.
- Corrected the **Xbox/HoloLens apps** row for Win32.
- Clarifying that UWP here means the Windows.UI.Xaml framework not UWP as a platform.
- Added **Uno Platform** as a cross-platform option for UWP and WinUI 3.
- Added a row for **Sandboxing**.
- Added additional star in the **Enterprise Apps** row for UWP because of sandboxing (many enterprise companies prefer sandboxed apps).

### DIFF
--- a/hub/apps/get-started/dev-options.md
+++ b/hub/apps/get-started/dev-options.md
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
 
 To read more about each of these Windows app development options, see [Writing apps for Windows](index.md).
 
-| Feature | .NET MAUI | React Native (RNW) | UWP | Win32 | Windows Forms | WinUI 3 | WPF |
+| Feature | .NET MAUI | React Native (RNW) | UWP XAML (Windows.UI.Xaml) | Win32 | Windows Forms | WinUI 3 | WPF |
 | --- | --- | --- | --- | --- | --- | --- | --- |
 | **Language** | C# | JavaScript, TypeScript | C#, C++, Visual Basic | C++, Rust | C#, Visual Basic | C#, C++ | C#, Visual Basic |
 | **UI language** | XAML/Code | JSX | XAML | Code | Code | XAML | XAML |
 | **UI designer**<br/>(drag & drop) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ | ✅ |
 | **UI debugging** | [Hot Reload](/dotnet/maui/xaml/hot-reload) | [Fast Refresh](https://reactnative.dev/docs/fast-refresh) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) | - | [Hot Reload](/visualstudio/debugger/hot-reload) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) |
-| **Fluent Design** | ✅ | ✅ | ❌ | ❌ | ❌ | ✅ | ❌ |
-| **.NET Runtime** | .NET | N/A | Minimal CLR runtime | N/A | .NET & .NET Framework | .NET | .NET & .NET Framework |
+| **Fluent Design** | ✅ | ✅ | ✅ (through [WinUI 2](/windows/apps/winui/winui2)) | ❌ | ❌ | ✅ | ❌ |
+| **.NET Runtime** | .NET | N/A | .NET Core & .NET Native | N/A | .NET & .NET Framework | .NET | .NET & .NET Framework |
 | **Windows App SDK** | ✅ ([more info](/dotnet/maui/platform-integration/invoke-platform-code)) | ✅ ([more info](https://techcommunity.microsoft.com/t5/modern-work-app-consult-blog/getting-started-with-react-native-for-windows/ba-p/912093)) | ❌ | ✅ | ✅ ([more info](../windows-app-sdk/winforms-plus-winappsdk.md)) | ✅ | ✅ ([more info](../windows-app-sdk/wpf-plus-winappsdk.md)) |
-| **Enterprise Apps** | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐⭐⭐ |
+| **Enterprise Apps** | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐⭐⭐ |
 | **Consumer Apps** | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐ |
 | **Great for touch** | ✅ | ✅ | ✅ | ❌ | ❌ | ✅ | ❌ |
-| **Cross-platform** | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ |
-| **Xbox/HoloLens apps** | ❌ | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ |
+| **Cross-platform** | ✅ | ✅ | ✅ (through [Uno Platform](/windows/apps/how-tos/uno-multiplatform)) | ❌ | ❌ | ✅ (through [Uno Platform](/windows/apps/how-tos/uno-multiplatform)) | ❌ |
+| **Xbox/HoloLens apps** | ❌ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |
+| **Sandboxing** | ❌ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |
 | **Currently supported** | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
 | **Receiving updates** | ✅ | ✅ | ✅ (security & bugfix) | ✅ | ✅ | ✅ | ✅ |
 | **Roadmap** | [GitHub](https://github.com/dotnet/maui/wiki/Roadmap) | [GitHub](https://aka.ms/rnw-roadmap) | n/a | n/a | [GitHub](https://github.com/dotnet/winforms/blob/main/docs/roadmap.md) | [GitHub](https://github.com/microsoft/WindowsAppSDK/blob/main/docs/roadmap.md) | [GitHub](https://github.com/dotnet/wpf/blob/main/roadmap.md) |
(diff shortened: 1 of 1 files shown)


jaigak:
Consider adding a section on app models separately and classify which frameworks run under UWP and which don't. Another thing to mention is that UWP XAML apps can only use an older version of .NET. Also add GDI or MFC in brackets for Win32.

ahmed605:
> Another thing to mention is that UWP XAML apps can only use an older version of .NET.

I mentioned that it's ".NET **Core**" which is an older version of .NET

jaigak:
> > Another thing to mention is that UWP XAML apps can only use an older version of .NET.
> 
> I mentioned that it's ".NET **Core**" which is an older version of .NET

Well yes, but .NET Core 2.2 was not the latest version of it.

alvinashcraft:
Hi @ahmed605. Thank you for your proposal. I am reviewing the suggestions with the product teams because the original document was created with input from each of the teams.

hub/apps/get-started/dev-options.md
```diff
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
...
-| Feature | .NET MAUI | React Native (RNW) | UWP | Win32 | Windows Forms | WinUI 3 | WPF |
+| Feature | .NET MAUI | React Native (RNW) | UWP XAML (Windows.UI.Xaml) | Win32 | Windows Forms | WinUI 3 | WPF |
 | --- | --- | --- | --- | --- | --- | --- | --- |
 | **Language** | C# | JavaScript, TypeScript | C#, C++, Visual Basic | C++, Rust | C#, Visual Basic | C#, C++ | C#, Visual Basic |
 | **UI language** | XAML/Code | JSX | XAML | Code | Code | XAML | XAML |
 | **UI designer**<br/>(drag & drop) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ | ✅ |
 | **UI debugging** | [Hot Reload](/dotnet/maui/xaml/hot-reload) | [Fast Refresh](https://reactnative.dev/docs/fast-refresh) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) | - | [Hot Reload](/visualstudio/debugger/hot-reload) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) |
-| **Fluent Design** | ✅ | ✅ | ❌ | ❌ | ❌ | ✅ | ❌ |
-| **.NET Runtime** | .NET | N/A | Minimal CLR runtime | N/A | .NET & .NET Framework | .NET | .NET & .NET Framework |
+| **Fluent Design** | ✅ | ✅ | ✅ (through [WinUI 2](/windows/apps/winui/winui2)) | ❌ | ❌ | ✅ | ❌ |
```
alvinashcraft:
```suggestion
| **Fluent Design** | ✅ | ✅ | ✅ (via [WinUI 2](/windows/apps/winui/winui2)) | ❌ | ❌ | ✅ | ❌ |
```
Just to shorten the text a bit

hub/apps/get-started/dev-options.md
```diff
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
...
+| **.NET Runtime** | .NET | N/A | .NET Core & .NET Native | N/A | .NET & .NET Framework | .NET | .NET & .NET Framework |
 | **Windows App SDK** | ✅ ([more info](/dotnet/maui/platform-integration/invoke-platform-code)) | ✅ ([more info](https://techcommunity.microsoft.com/t5/modern-work-app-consult-blog/getting-started-with-react-native-for-windows/ba-p/912093)) | ❌ | ✅ | ✅ ([more info](../windows-app-sdk/winforms-plus-winappsdk.md)) | ✅ | ✅ ([more info](../windows-app-sdk/wpf-plus-winappsdk.md)) |
-| **Enterprise Apps** | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐⭐⭐ |
+| **Enterprise Apps** | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐⭐⭐ |
 | **Consumer Apps** | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐ | ⭐⭐⭐ | ⭐⭐⭐⭐⭐ | ⭐⭐⭐⭐ |
 | **Great for touch** | ✅ | ✅ | ✅ | ❌ | ❌ | ✅ | ❌ |
-| **Cross-platform** | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ | ❌ |
-| **Xbox/HoloLens apps** | ❌ | ✅ | ✅ | ✅ | ❌ | ❌ | ❌ |
+| **Cross-platform** | ✅ | ✅ | ✅ (through [Uno Platform](/windows/apps/how-tos/uno-multiplatform)) | ❌ | ❌ | ✅ (through [Uno Platform](/windows/apps/how-tos/uno-multiplatform)) | ❌ |
```
alvinashcraft: Avalonia UI could also be mentioned for WPF. However, the team does not want to bring these 3rd party options into this page yet. We may revisit this decision later. We're not opposed to discussion the benefits of Uno Platform, and we have a couple of pages about using Uno to take WinUI 3 code cross-platform. Thanks for raising this though.

hub/apps/get-started/dev-options.md
```diff
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
...
+| **Xbox/HoloLens apps** | ❌ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |
```
alvinashcraft, line 33: This is a good change. Thanks for this fix.

hub/apps/get-started/dev-options.md
```diff
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
...
+| **Sandboxing** | ❌ | ✅ | ✅ | ❌ | ❌ | ❌ | ❌ |
```
alvinashcraft: The team doesn't want to add this row here. It's something that can be confused with other types of security options for app deployment. For example, the ability to run most apps in a lightweight appContainer if they're MSIX packaged and deployed with TrustLevel set to appContainer: https://learn.microsoft.com/windows/msix/desktop/desktop-to-uwp-behind-the-scenes#types-of-desktop-app

hub/apps/get-started/dev-options.md
```diff
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
...
+| Feature | .NET MAUI | React Native (RNW) | UWP XAML (Windows.UI.Xaml) | Win32 | Windows Forms | WinUI 3 | WPF |
 | --- | --- | --- | --- | --- | --- | --- | --- |
 | **Language** | C# | JavaScript, TypeScript | C#, C++, Visual Basic | C++, Rust | C#, Visual Basic | C#, C++ | C#, Visual Basic |
 | **UI language** | XAML/Code | JSX | XAML | Code | Code | XAML | XAML |
 | **UI designer**<br/>(drag & drop) | ❌ | ❌ | ✅ | ❌ | ✅ | ❌ | ✅ |
 | **UI debugging** | [Hot Reload](/dotnet/maui/xaml/hot-reload) | [Fast Refresh](https://reactnative.dev/docs/fast-refresh) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) | - | [Hot Reload](/visualstudio/debugger/hot-reload) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) | [Hot Reload](/visualstudio/xaml-tools/xaml-hot-reload) |
-| **Fluent Design** | ✅ | ✅ | ❌ | ❌ | ❌ | ✅ | ❌ |
-| **.NET Runtime** | .NET | N/A | Minimal CLR runtime | N/A | .NET & .NET Framework | .NET | .NET & .NET Framework |
+| **Fluent Design** | ✅ | ✅ | ✅ (through [WinUI 2](/windows/apps/winui/winui2)) | ❌ | ❌ | ✅ | ❌ |
+| **.NET Runtime** | .NET | N/A | .NET Core & .NET Native | N/A | .NET & .NET Framework | .NET | .NET & .NET Framework |
```
alvinashcraft: This was set to "Minimal CLR runtime" because the runtime for .NET Native is MRT.dll (Minimal CLR Runtime): https://learn.microsoft.com/windows/uwp/dotnet-native/

hub/apps/get-started/dev-options.md
```diff
@@ -17,20 +17,21 @@ There is a wide range of options for developing applications for Windows. The be
 
 To read more about each of these Windows app development options, see [Writing apps for Windows](index.md).
 
-| Feature | .NET MAUI | React Native (RNW) | UWP | Win32 | Windows Forms | WinUI 3 | WPF |
+| Feature | .NET MAUI | React Native (RNW) | UWP XAML (Windows.UI.Xaml) | Win32 | Windows Forms | WinUI 3 | WPF |
```
alvinashcraft, line 20: This is tricky, but the team is willing to accept this change and see how it's received. There isn't an easy way to compare these options and pick out just the "UI framework" part of each one. .NET MAUI is much more than a UI framework.

alvinashcraft (changes requested):

hub/apps/get-started/dev-options.md
ahmed605: Should it be changed to "Sandboxing (AppContainer)" instead?
WinUI 3 (and other UI frameworks) don't officially support `TrustLevel="appContainer"` yet, you would get crashes on it, so I think it makes sense to add this row to declare that it's only officially supported for UWP to prevent devs from using it on other UI frameworks and getting crashes as a result.

alvinashcraft: We could try that and see how it's received. I think that makes it a bit clearer.

hub/apps/get-started/dev-options.md
ahmed605: It's minimal as in it's a minimal implementation but the API surface is actually almost the whole .NET Core 2.2 API surface (+ additional stuff even), so mentioning that it's minimal is kinda misleading for devs.
(this was mentioned in the Channel9's [Inside .NET Native](https://web.archive.org/web/2020/https://channel9.msdn.com/Shows/Going+Deep/Inside-NET-Native) video, they mention there that it's the same APIs but with minimal/optimized implementation, the impl also became better in the final release of ProjectN (the video was about the Technical Preview version))

hub/apps/get-started/dev-options.md
ahmed605: That's true, but the problem is that there's rows for UI-Framework-specific things (e.g the designer and Fluent Design) and these don't apply to non-XAML UWP apps like DX, Composition, or Flutter ones

hub/apps/get-started/dev-options.md
alvinashcraft: So, maybe we should rename the row to ".NET API Surface" or ".NET variation". I want to say "type" but that word is misleading too.

ahmed605: I think ".NET API Surface" or just ".NET" would work, I would prefer the latter.

alvinashcraft: OK. Let's try that and see how it's received by others in the community.

alvinashcraft:
I also wanted to let you know that we're discussing the use of star ratings. The rating makes it hard to tell if we're saying that it's "recommended" or "popular" or "good at this".

alvinashcraft (approved):
This looks good to me now. I'll merge this after the build succeeds. Thanks for the contribution!

hub/apps/get-started/dev-options.md
jaigak: > This is tricky, but the team is willing to accept this change and see how it's received. There isn't an easy way to compare these options and pick out just the "UI framework" part of each one. .NET MAUI is much more than a UI framework.

It should not be tricky. 'UWP' should only refer to the sandboxed application model (CoreApplication) and it can run multiple UI frameworks on top. Unfortunately, Microsoft has had a long history of using confusing and conflated terminology.

PR #4694, opened 2023-10-24
Update with info on cloning an existing projects in VS

Mikejo5000:
Minor edit based on new support for cloning an existing repo with Dev Drive in VS.


mattwojo:
Thanks @Mikejo5000! Merged.

PR #4708, opened 2023-10-31
Update setup.md -- GA of Dev Drive

schmik:
Removed warning considering availability of Dev Drive. Dev Drive is now GA: https://devblogs.microsoft.com/engineering-at-microsoft/dev-drive-is-now-available/


mattwojo:
Good catch! Thanks @schmik! I had removed that Insiders note on the Dev Drive, but missed it here. Appreciate the assist!

PR #4710, opened 2023-11-01
Remove extra ".com" in the hyperlink

jerobado:


mattwojo:
Good catch. Thank you @jerobado!

PR #4743, opened 2023-11-21
[Accessibility] Document F6!

citelao:
The **F6 key** is a critical keyboard usability feature for many users, especially keyboard-only users who do not use screen readers. However, we do not document it today. This change adds some _basic_ F6 documentation to our accessibility & keyboarding guidelines.

## What changed?

* Added docs to Keyboard accessibility page
* Added docs to Landmarks & headings page
* Added docs (and image) to Keyboard interactions page

## Justification

F6 is a well-known shortcut. In fact, we publicly document it in [Keyboard shortcuts in Windows](https://support.microsoft.com/en-gb/windows/keyboard-shortcuts-in-windows-dcc61a57-8ff0-cffe-9796-cb9706c75eec#:~:text=F6,on%20the%20desktop.). 

It is *critical* to the usability of several of key surfaces, including Edge, the literal Windows desktop, File Explorer, and many Office products (including PowerPoint and Outlook). It is a rapid accelerator for navigating the UI with keyboard (and without a screen reader, which has other tools).

However, support for F6 is declining. It is rare to find F6 support in newer apps, which instead tend to rely on hard-to-find hotkeys or, worse, just tab and arrow keys.

This is partially a documentation issue. We should document F6 as an important shortcut, and then we can discuss how to introduce F6 support more consistently in our UI.


citelao:
We might want to add to the [list of Common Keyboard Accelerators](https://learn.microsoft.com/en-us/windows/apps/design/input/keyboard-accelerators#common-keyboard-accelerators)

citelao:
@Karl-Bridge-Microsoft, anything blocking this?

citelao:
Thank you for reviewing this! I'm so excited to reference this content in the future.

PR #4783, opened 2023-12-13
[Attached properties] Remove incorrect documentation about C++/WinRT

citelao:
Today, our documentation incorrectly claims that custom attached properties cannot be set in XAML markup if they were implemented in C++/WinRT. This change updates our documentation to correct the mistake.

## What changed?

* Removed C++/WinRT callouts
* Added note about this change, since it is subtle.

## Why?

We tested this locally; we have a custom attached property & it works just fine in XAML markup ([internal PR link](https://microsoft.visualstudio.com/OS/_git/OSClient/pullrequest/9998329)).

### DIFF
--- a/uwp/xaml-platform/custom-attached-properties.md
+++ b/uwp/xaml-platform/custom-attached-properties.md
@@ -12,6 +12,7 @@ dev_langs:
   - cppwinrt
   - cpp
 ---
+
 # Custom attached properties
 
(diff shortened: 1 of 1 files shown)


stevewhims:
I'll take this. The topic is assigned to Jim, but I added this content (back when it was true). :)

stevewhims:
I've removed the bit saying that the topic was previously incorrect. We update content all the time when the platform changes (as in this case), and we make corrections to content all the time. So if we left "this topic was previously wrong about" messages each time we did that, the topic would say little else. :)

citelao:
Thanks, @stevewhims! I wasn't sure if it had changed or if it was a doc bug. Your edits are fine :).

PR #4795, opened 2023-12-20
Fix various typos

Blake-Madden:


hub/dev-environment/javascript/nodejs-overview.md
```diff
@@ -31,7 +31,7 @@ Node.js is primarily used for building fast and scalable web applications. It us
...
+- **Server-side rendered apps (SSRs)**: These web apps can run on both the client (in your browser / the front-end) and the server (the back-end) allowing pages that are dynamic to display (generate HTML for) whatever content is known and quickly grab content that is not known as it's available. These are often referred to as “isomorphic” or “universal” applications. SSRs utilize SPA methods in that they don't need to reload every time you use it. SSRs, however, offer a few benefits that may or may not be important to you, like making content on your site appear in Google search results and providing a preview image when links to your app are shared on social media like Twitter or Facebook. The potential drawback being that they require a Node.js server constantly running. In terms of examples, a social networking app that supports events that users will want to appear in search results and social media may benefit from SSR, while an email app may be fine as an SPA. You can also run server-rendered no-SPA apps, which may be something like a WordPress blog. As you can see, things can get complicated, you just need to decide what's important.
```
alvinashcraft:
```suggestion
- **Server-side rendered apps (SSRs)**: These web apps can run on both the client (in your browser / the front-end) and the server (the back-end) allowing pages that are dynamic to display (generate HTML for) whatever content is known and quickly grab content that is not known as it's available. These are often referred to as "isomorphic" or "universal" applications. SSRs utilize SPA methods in that they don't need to reload every time you use it. SSRs, however, offer a few benefits that may or may not be important to you, like making content on your site appear in Google search results and providing a preview image when links to your app are shared on social media like Twitter or Facebook. The potential drawback being that they require a Node.js server constantly running. In terms of examples, a social networking app that supports events that users will want to appear in search results and social media may benefit from SSR, while an email app may be fine as an SPA. You can also run server-rendered no-SPA apps, which may be something like a WordPress blog. As you can see, things can get complicated, you just need to decide what's important.
```
Fixing the quotation marks.

alvinashcraft (approved):
Thanks for all these fixes. Looks good!

PR #4796, opened 2023-12-20
Update color-picker.md

NathanD-GDS:
This PR makes a variety of improvements to the [PowerToys Color Picker documentation](https://learn.microsoft.com/en-us/windows/powertoys/color-picker). These improvements include:

* aligning the page with the Microsoft style guide, specifically:
  * [how mouse actions are described](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/term-collections/mouse-mouse-interaction-terms#mouse-actions) (hover, select, scroll)
  * [use of contractions](https://learn.microsoft.com/en-us/style-guide/word-choice/use-contractions)
  * [how toggles are described](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/t/toggle)
  * [use active voice](https://learn.microsoft.com/en-us/style-guide/grammar/verbs#active-and-passive-voice)
* changed verbs to better suit the utility name (especially changing 'Select' to 'Pick')
* removed some duplication around **Color formats** settings
* made clipboard references consistent ('the clipboard', not 'your clipboard')
* simplifies some of the page's language

### DIFF
--- a/hub/powertoys/color-picker.md
+++ b/hub/powertoys/color-picker.md
@@ -31,7 +31,7 @@ You can choose what happens when you activate Color Picker (default: <kbd>Win</k
 
 ### Picking colors
 
-After activating Color Picker, hover over a color and click to pick that color. If you want to see the area under your cursor in more detail, scroll up to zoom in.
+After activating Color Picker, select a color on your screen to pick that color. If you want to see the area under your cursor in more detail, scroll up to zoom in.
 
 Color Picker copies the selected color to the clipboard in the **Default color format** you've chosen in Color Picker's settings (default: HEX).
 
@@ -61,7 +61,7 @@ To choose a similar color, select one of the segments on the left and right edge
 
 To remove a color from the colors history, right-click a color and select **Remove**.
 
-To export the colors history, right-click a color and hover over **Export**. You can group the exported values by colors or formats.
+To export the colors history, select and hold (or right-click) a color and select **Export**. You can group the exported values by colors or formats.
 
 ## Settings
 
--- a/hub/powertoys/color-picker.md
+++ b/hub/powertoys/color-picker.md
@@ -1,6 +1,6 @@
 ---
 title: PowerToys Color Picker utility for Windows
-description: A system-wide color picking utility for Windows to pick colors from the screen and copy the default value to your clipboard.
+description: A system-wide color picking utility for Windows to pick colors from the screen and copy the default value to the clipboard.
 ms.date: 08/03/2023
 ms.topic: article
 ms.localizationpriority: medium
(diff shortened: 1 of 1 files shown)


hub/powertoys/color-picker.md
```diff
@@ -9,31 +9,31 @@ no-loc: [PowerToys, Windows, ColorPicker, Color, Picker]
...
-- **Only pick a color** - Opens Color Picker only and the selected color will be copied to the clipboard.
+- **Open editor** opens an editor that lets you choose a color from the colors history, fine-tune a selected color, or pick a new color
+- **Pick a color and open editor** activates Color Picker, then opens an editor and copies the selected color to the clipboard after you've picked a color
+- **Only pick a color** activates Color Picker and copies the selected color to the clipboard
 
-### Selecting colors
+### Picking colors
 
-After the Color Picker is activated, hover your mouse cursor over the color you would like to copy and left-click the mouse button to select a color. If you want to see the area under your cursor in more detail, scroll up to zoom in.
+After activating Color Picker, hover over a color and click to pick that color. If you want to see the area under your cursor in more detail, scroll up to zoom in.
```
alvinashcraft: This is an improvement over the original sentence, but it would be better to avoid the work "click". I may be on a touch device, tapping the controls. That's why we typically use "select" instead of "click".

hub/powertoys/color-picker.md
```diff
@@ -46,45 +46,52 @@ The copied color will be stored in your clipboard in the format that is configur
...
+
+To choose a similar color, select one of the segments on the left and right edges of the color bar. The Color Picker editor suggests two lighter shades on the left of the bar, and two darker shades on the right of the bar. Selecting one of these similar colors adds that color to the colors history.
 
 ![ColorPicker Editor window.](../images/pt-colorpicker-editor.gif)
 
-To remove any color from the colors history, right-click a color and select **Remove**.
+To remove a color from the colors history, right-click a color and select **Remove**.
 
-To export the colors history, right-click a color and select **Export**. The exported values can be grouped by colors or formats.
+To export the colors history, right-click a color and hover over **Export**. You can group the exported values by colors or formats.
```
alvinashcraft: This is another case where touch input isn't accounted for. "Hover over" isn't possible when using touch. I would either leave "select" or do something like "hover over or tap"... Select is probably the simplest option, unless you have another idea.

alvinashcraft (commented):
Thanks for all the improvements. It's much easier to read. I left a couple of minor suggestions. Let me know what you think.

hub/powertoys/color-picker.md
NathanD-GDS: Agreed! I've committed a change for this now. Let me know what you think.

hub/powertoys/color-picker.md
NathanD-GDS: Good point. I'd followed some examples in [the style guide](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/r/right-click) in my new commit.

alvinashcraft (approved):
The new changes look good. Thanks again!

PR #4812, opened 2024-01-07
Update manifest URLs for winget docs

rakleed:
Since winget-cli currently uses manifest version 1.5.0.


mattwojo:
@denelon - Could you confirm this update?

denelon (approved):

denelon:
Yes, and we're close to 1.6.  I'll try to include a note about checking these docs in the future releases.

mattwojo:
Thanks @denelon. Yes, if there is a way to build a docs update into the release plan, that would be super helpful!
And thank you @rakleed for contributing this update!

PR #4820, opened 2024-01-13
Makes headings consistent across PowerToys documentation

NathanD-GDS:
The headings across the PowerToys documentation are very inconsistent.

This PR:

* makes headings more consistent by applying the most used verb form in the PowerToys docs - the present participle - across as many pages and headings as possible (`Use PowerToys Run` becomes `Using PowerToys Run`)
* makes non action-based headings, action-based (for example, `Plugin manager` becomes `Using plugins for PowerToys Run`)
* makes generic headings specific to the utility (`How to use` becomes `Using Quick Accent`)
* simplifies the heading structure on some pages
* improves the heading structure of pages that had too broad of a heading structure
* adds missing `utility` suffix to H1s
* makes short, undescriptive headings more specific (for example, `Limitations` becomes `Limitations of Color Picker`

Some generic headings, such as `Settings`, have been left alone.

If it turns out that another verb format is better for headings, this PR should make it easier to change as most headings now follow a similar structure.

### DIFF
--- a/hub/powertoys/always-on-top.md
+++ b/hub/powertoys/always-on-top.md
@@ -12,7 +12,7 @@ A system-wide Windows utility to pin windows above other windows.
 
 ![AlwaysOnTop screenshot.](../images/pt-always-on-top.png)
 
-## Pin a window
+## Pinning a window
 
 When you activate Always On Top (default: <kbd>⊞ Win</kbd>+<kbd>Ctrl</kbd>+<kbd>T</kbd>), the utility pins the active window above all other windows. The pinned window stays on top, even when you select other windows.
 
--- a/hub/powertoys/awake.md
+++ b/hub/powertoys/awake.md
@@ -10,7 +10,7 @@ ms.localizationpriority: medium
 
 PowerToys Awake is a tool for Windows designed to keep a computer awake without having to manage its [power & sleep settings](https://support.microsoft.com/windows/how-to-adjust-power-and-sleep-settings-26f623b5-4fcc-4194-863d-b824e5ea7679). This can be helpful when running time-consuming tasks, ensuring that the computer does not go to sleep or turn off its screens.
 
-## Getting started
+## Using PowerToys Awake
 
 You can use PowerToys Awake directly from PowerToys Settings or as a standalone executable. When it's running from PowerToys, it can be managed from PowerToys Settings or the system tray.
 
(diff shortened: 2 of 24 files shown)


alvinashcraft:
Thanks for the submission, @NathanD-GDS. I'll have to discuss this with the team on our call this week. I try to avoid changing H1s and H2s unless they're incorrect because folks out there may have bookmarked or shared links that jump to those specific sections. We may have cross-links in the docs that also do this.

NathanD-GDS:
> Thanks for the submission, @NathanD-GDS. I'll have to discuss this with the team on our call this week. I try to avoid changing H1s and H2s unless they're incorrect because folks out there may have bookmarked or shared links that jump to those specific sections. We may have cross-links in the docs that also do this.

That makes sense, @alvinashcraft . I hadn't considered the possibility of bookmarks. I think I did a cursory look for cross-links but I wasn't exhaustive. Understand if this PR is a no-go.

htcfreek:
@alvinashcraft , @NathanD-GDS 
For the policy page there is another reason to not change them: The header of the policies is the policy name.

alvinashcraft:
[like]  Alvin Ashcraft reacted to your message:
________________________________

alvinashcraft:
Good point, @htcfreek.

@NathanD-GDS I don't want to discard the PR because I think there are some good changes here. Let's wait until several other PRs merge with the next release in a few days and then do the following:

- Merge the latest changes from other PRs to your branch and resolve any conflicts
- Exclude the changes to the Group Policy page because the headers need to match the policy names
- Add anchor tags to the updated header text to preserve the original header links. Here's an example:

`## <a id="anchortext">Header text</a>`

Let me know what you think. I know it adds a bit of work to implement the last suggestion.

Jay-o-Way:
I too appreciate the effort. Few valid comments up above here. After a while, there's no shame in starting over with a fresh branch.
(speaking from experience:)

Jay-o-Way:
Ping @crutkas

crutkas:
looking at aka links that i control, here are the two i spotted with anchors.  These would need to have backup anchors

https://learn.microsoft.com/windows/powertoys/keyboard-manager#keys-that-cannot-be-remapped
https://learn.microsoft.com/windows/powertoys/mouse-utilities#mouse-jump

alvinashcraft:
Following up on this one from a year ago. After the discussing here, and given how much things of changed in the meantime, I think any changes in this area should be in the scope of a new PR. I'm closing this now but will monitor for any further replies. Thanks again for the work put into this one.

PR #4826, opened 2024-01-17
Fix minor typo

tolache:
Replace `-scope` with `-scope`


mattwojo:
Thanks @tolache! Great catch!

PR #4831, opened 2024-01-20
PowerToys > Policy docs: Remove part for "Auto update" policy

htcfreek:
We decided to never implement this policy. Therefore the prepared text becomes obsolete.

xref: microsoft/PowerToys/issues/24363

This can auto-merge immediately.


alvinashcraft (approved):
Looks good!

PR #4833, opened 2024-01-22
Fix some minor mistakes

PesBandi:
Changed the capitalization of some names so that it matches the PowerToys settings, fixed a typo, updated the command for Paste As Plain Text and moved PT Run before Quick Accent so that they are sorted alphabetically.


alvinashcraft (approved):
Looks good. Thanks for the clean-up!

PR #4863, opened 2024-02-12
Update quickstart-save-a-file-with-a-picker.md

andreas-tornqvist:
In current description, the name of the picked file goes where the contents should be, when writing the file


alvinashcraft (approved):
Thank you for fixing this code, @andreas-tornqvist. This looks good.

PR #4889, opened 2024-03-03
Update manifest URLs for winget docs

rakleed:
Since winget-cli currently uses manifest version 1.6.0.


mattwojo:
Thanks for these updates @rakleed ! Merged.  🥇

Issue #3630, opened 2022-01-18
React Tutorial Needs Updating


TWebSrDev:
[Enter feedback here]
The "React on Windows for beginners" tutorial is woefully out-of-date. It teaches React development using classes and component lifecycle methods, when it should be teaching React dev with functional components and hooks. For the last few years, there has been a major shift in React towards functional components and Hooks, and this tutorial needs to updated to reflect modern React app architecture and development techniques. This tutorial is akin to teaching web development using VB and web forms.
---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 945dd697-a98b-1d9b-28b0-6f81b5122cf3
* Version Independent ID: ff53c8e3-ac95-d022-a9d5-cab9b933960e
* Content: [Tutorial on React for beginners](https://docs.microsoft.com/en-us/windows/dev-environment/javascript/react-beginners-tutorial)
* Content Source: [hub/dev-environment/javascript/react-beginners-tutorial.md](https://github.com/MicrosoftDocs/windows-uwp/blob/docs/hub/dev-environment/javascript/react-beginners-tutorial.md)
* Product: **dev-environment**
* Technology: **windows-javascript**
* GitHub Login: @mattwojo
* Microsoft Alias: **mattwoj**

mattwojo:
Thanks for filing @TWebSrDev. Agree, this should be updated. Does everything look ok to you up to the "Try using React with an API" section? 
https://docs.microsoft.com/en-us/windows/dev-environment/javascript/react-beginners-tutorial#try-using-react-with-an-api 

I'll set aside some time soon to try a re-write, but I'm also no React expert... if you'd like to take a crack at a re-write, please feel welcome to submit as a PR. Otherwise I'll link my commit once I have a chance to take a crack at it.

TWebSrDev:
Thanks for the response @mattwojo!
Ha! I'm no React expert either, but know my way around it a bit. 😀 

Yes, what I read up to that point looked good. In fact, I really liked the doc, and think that it's a great beginner's tutorial because it actually teaches something immediately applicable to "real world", as opposed to the typical "Todo List"-type tutorial.

[...]

PR #4903, opened 2024-03-13
Update Javascript React related documentaion

vimode:
Hello, 

As per discussion in #3630 I have started updating the documentation related to React. 
I am still working on, and updating the tutorials, the old React related content wherever necessary, so the draft PR. 

@mattwojo you can start reviewing the changes whenever you can.

There is some content and  resources which still links to the old docs as the new docs do not have that information, should it be removed and/or replaced with some other links or as long as they are live its good to have them the same?

Thank you.


vimode:

mattwojo:
Looks good. Thank you @vimode !!
Merged. 🥇

PR #4921, opened 2024-03-24
Fix misspellings and other typos

Blake-Madden:


uwp/networking/background-transfers.md
```diff
@@ -76,7 +76,7 @@ Note the asynchronous method calls defined using JavaScript promises. Looking at
 promise = upload.startAsync().then(complete, error, progress);
 ```
 
-The async method call is followed by a then statement which indicates methods, defined by the app, that are called when a result from the async method call is returned. For more information on this programming pattern, see [Asynchronous programming in JavaScript using promises](/previous-versions/windows).
```
Jay-o-Way, line 79: In this case, `then` is actually code. So it must be there, maybe format as code.
```cs
  promise = upload.startAsync().then(complete, error, progress);
```

Jay-o-Way (changes requested):
Nice work! Just one thing...

jwmsft:
Thanks for submitting these corrections!

jwmsft (approved):

Issue #4936, opened 2024-04-07
Potentially wrong date on WSA Deprecation notice. 


hrichiksite:
### Type of issue

Typo

### Feedback

https://learn.microsoft.com/en-us/windows/android/wsa/#:~:text=Customers%20that%20have,we%20evolve%20experiences.

The Notice mentions 
> Customers that have installed the Amazon Appstore or Android apps prior to **March 5, 2024**, will continue to have access to those apps through the deprecation date of March 5, 2025.

The first date here (March 5, 2024) may be wrong, I think it was meant to be March 5, 2025.

The same is written here on Amazon Appstore's page.
https://developer.amazon.com/apps-and-games/appstore-on-windows-11

### Page URL

https://learn.microsoft.com/en-us/windows/android/wsa/

### Content source URL

https://github.com/MicrosoftDocs/windows-dev-docs/blob/docs/hub/android/wsa/index.md

### Author

@mattwojo

### Document Id

31f9a801-c5b5-94e7-41b9-1b2b4a1d27de

mattwojo:
This isn't a mistake. The end of support announcement was posted in March 2024, but as the notice states:

"Customers that have installed the Amazon Appstore or Android apps prior to March 5, 2024, will continue to have access to those apps through the deprecation date of March 5, 2025."

PR #4937, opened 2024-04-07
Fix potentially wrong date on WSA deprecation docs.

hrichiksite:
Fixes (if a problem at all) https://github.com/MicrosoftDocs/windows-dev-docs/issues/4936


mattwojo:
Not a mistake. But thank you for the question and offer to contribute a fix. 

"Customers that have installed the Amazon Appstore or Android apps prior to March 5, 2024, will continue to have access to those apps through the deprecation date of March 5, 2025."

PR #4941, opened 2024-04-08
Remove Unix directory separator

ofthelit:
Use the same spelling of `C: drive` everywhere.


mattwojo:
Thanks for the contribution @ofthelit !

PR #4944, opened 2024-04-10
Update download.md - Fixed code typo.

smtpcx:
The last code example on this page shows how to download an installer to a specific folder location, but the example incorrectly specifies 'winget install' instead of 'winget download'.


mattwojo:
Good catch @smtpcx . Thank you for the fix!

PR #4935, opened 2024-04-06
formatting and small text adjustments

Jay-o-Way:
Did a rather general pass on these docs to improve some formatting. Also adjusted some text here and there, but all in small changes.
Changed a number of "launch" words to "open", but excluded the words that are still in the UI.

Question for a (possible) future PR: I see both "powertoys-" and "pt-" prefixes are used in the file names for images. Should we use one and the same? And if so, which one?

PR #4946, opened 2024-04-11
Formatting and phrases

Jay-o-Way:
Pretty similar to #4935.
A _little bit_ related to https://github.com/microsoft/PowerToys/pull/32351/files. Release 0.81 is expected around MS Build (late May) but as there are only a handful changes which are in **both** the UI **and** in the Docs, I don't see a problem changing this now.


Jay-o-Way:
FYI @ethanfangg

hub/powertoys/general.md
```diff
@@ -38,4 +36,4 @@ Set a location where you want to save your PowerToys settings. You can also rest
 
 ## Experimentation
 
-Will activate experimentation with new features for Windows Insider builds.
+Will activate experimentation with new features on Windows Insider builds, if availabe.
```
alvinashcraft:
```suggestion
Will activate experimentation with new features on Windows Insider builds, if available.
```

hub/powertoys/install.md
```diff
@@ -132,6 +132,6 @@ After successfully installing PowerToys, an overview window will display with in
 
 ## Updates
 
-PowerToys uses an auto-updater that checks for new versions when the app is running. If enabled, a toast notification will appear when an update is available. Updates can also be checked for manually from the PowerToys Settings, under the General page.
+PowerToys uses an auto-updater that checks for new versions when the app is running. If enabled, a toast notification will appear when an update is available. Updates can also be checked for manually from the PowerToys Settings, on the **General** page.
```
alvinashcraft: In some previous docs in this PR, I noticed that the page name in settings was not bold. I'll have to go back and find those, if you want to make them consistent. Checking for those could probably be another PR though.

hub/powertoys/run.md
```diff
@@ -241,8 +241,8 @@ Examples:
 
 ### Unit converter plugin
 
-> [!NOTE]
-> The Unit Converter plugin respects the number format settings of your system. Please be aware of the different decimal and thousand delimiters in different locals. The names and abbreviations of the units aren't localized yet.
+> [!IMPORTANT]
+> The Unit Converter plugin respects the number format settings of your system. Please be aware of the different decimals and thousand delimiters in different locals. The names and abbreviations of the units aren't localized yet.
```
alvinashcraft: I think decimals should remain decimal. It's talking about "decimal delimiters" and "thousand delimiters".

alvinashcraft (commented):
Looks good overall. Just a couple of minor comments/suggestions. Thanks for going through all these pages.

hub/powertoys/run.md
Jay-o-Way: Yeah, subtle but tricky. "Decimal delimiters" doesn't sound right. There can be only one decimal char, so it can't be delimited. Another word could be "_thousand(s)_ separator". Might be best to split this into "decimal character and thousands-separator".

alvinashcraft: Yes, I like your suggestion to break it up.

hub/powertoys/install.md
Jay-o-Way: Yeah in a couple of places i just removed this all together. It feels stupid to tell so-called "power users" to go to a certain page in a Settings app, to change certain settings.

Jay-o-Way:
done :)

alvinashcraft (approved):
Looks great. Thanks again!

PR #4966, opened 2024-04-23
#32527 - Running with Administrator Permissions documentation is missing Mouse without Borders

deanis74:
Added Mouse without Borders to Affected PowerToys Utilities. Also made two small changes to fix the alpha sort of the PowerToys in the list.


alvinashcraft (approved):
Thanks for adding this (and for fixing the order). Looks good.

Issue #2576, opened 2020-07-13
Mention QueryOptionalDelayLoadedAPI in Version Adaptive App docs


dostende:
I think this documentation should provide a link to `QueryOptionalDelayLoadedAPI` as well. I wanted to check for and dynamically use `DXGIDeclareAdapterRemovalSupport` but any search for "uwp check if api available" or similar lead me to the Version Adaptive App docs instead.

---
#### Document Details

⚠ *Do not edit this section. It is required for docs.microsoft.com ➟ GitHub issue linking.*

* ID: 179cd563-b3b3-10c5-709d-87d81835850f
* Version Independent ID: 890ebf14-903b-6678-c785-37c8a4c8f868
* Content: [Version adaptive apps - UWP applications](https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/version-adaptive-apps)
* Content Source: [windows-apps-src/debug-test-perf/version-adaptive-apps.md](https://github.com/MicrosoftDocs/windows-uwp/blob/docs/windows-apps-src/debug-test-perf/version-adaptive-apps.md)
* Product: **uwp**
* Technology: **deployment**
* GitHub Login: @Karl-Bridge-Microsoft
* Microsoft Alias: **kbridge**

Karl-Bridge-Microsoft:
Thanks for the suggestion. Topic has been updated.

PR #4971, opened 2024-04-27
Fix a broken url

ooooo-create:
ref [#2576](https://github.com/microsoft/devhome/pull/2576)

The `docs/extensions.md` have changed to `docs/extensions/readme.md`.


mattwojo:
Thanks for the fix @ooooo-create !

PR #4973, opened 2024-04-28
Update config.yml

arseniigorkin:
edited http typo


mattwojo (approved):
Great catch - thank you!

PR #4982, opened 2024-05-07
Fix XML in desktop-to-uwp-extensions.md

MikeHillberg:
Syntax error in XML, and note that it's `uap:` not `uap2:`

https://learn.microsoft.com/en-us/uwp/schemas/appxpackage/uapmanifestschema/element-uap-supportedfiletypes


stevewhims (approved):

stevewhims:
I checked across the docs; the only other typos along these lines were in the same file, so I added those fixes to this PR.

PR #4985, opened 2024-05-08
Add missing quotes on command

MatthewSteeples:
These quotes are necessary to run the command


mattwojo:
Thanks @MatthewSteeples ! I was able to add a single filter without using quotes, but looks like multiple does require quotes. Appreciate the catch!

PR #4987, opened 2024-05-10
Update react-on-windows.md

naragon:
The first sentence incorrectly stated that this guide is for React on WSL when it describes the process for plain Windows OS.


mattwojo (approved):
Good catch - thank you for the fix!

PR #4993, opened 2024-05-14
Removed info that said that a Dev Drive needs a Windows Insider Program Build on the Dev Channel.

LinusNoren:
Removed info that said that a Dev Drive needs a Windows Insider Program Build on the Dev Channel.


mattwojo:
Good catch. Thank you!!

PR #5003, opened 2024-05-21
changing message for mapping without azure portal

jomorr:
Azure portal is being decommissioned, so altering public instructions for requesting the PFN mapping


jomorr:
#sign-off

PR #5004, opened 2024-05-21
Update quick-accent.md

timmy-wright:
Add macron to Māori.


alvinashcraft (approved):
Looks good. Thanks for fixing this!

PR #5005, opened 2024-05-23
Update advanced-paste.md to include OpenAI account requirement

D4n2021:
added OpenAI account requirement


alvinashcraft (approved):
Good to call it out up here. Thanks!

PR #5009, opened 2024-05-24
Update recommended code editor from Intellij to WebStorm

Bofutw:
Changed the recommended code editor from IntelliJ IDEA to WebStorm, which is more suitable for developers working with Node.js.


mattwojo:
Good recommendation. Thanks for the update!

PR #5023, opened 2024-06-04
Fix video embed for Powertoys intro page

jongalloway:
This video embed isn't working, docs indicate these should be prefaced by `> ` to work.


alvinashcraft (approved):
Thanks for the fix!

PR #5029, opened 2024-06-13
Update custom-dependency-properties.md to provide example and more details of proper collection initialization

michael-hawker:
This is a common pitfall that I myself have made but that I see others make as well.

As by default you'll initialize it to null...
Then get an error about XAML not being able to add to an uninitialized collection...
So then folks naturally set the default to a new collection...
Which would make sense, except that's not what needs to done as it's static so it then shares that collection across instances.

This is called out in the docs here, but it's not immediate evident that's what this scenario is describing and what it's talking about without an example.

This PR adds the common pattern that's what not to do in contrast with the proper fix as well as more details about what's happening.


jwmsft (approved):

jwmsft:
@michael-hawker, thank you!

PR #5030, opened 2024-06-13
Update launch-settings-app.md

TheOnlyOnlyName:
Added dynamic lighting settings URI


alvinashcraft (approved):
Thanks for adding this URI. Tested it locally and it works great.

PR #5035, opened 2024-06-20
Fix a few typos

Blake-Madden:


GrantMeStrength:
Thank you! Even though code nippets and toogle switches should really be a thing.

PR #5060, opened 2024-07-10
Update drag-and-drop.md

asklar:
Add a mention about setting background to _something_

### DIFF
--- a/hub/apps/design/input/drag-and-drop.md
+++ b/hub/apps/design/input/drag-and-drop.md
@@ -52,7 +52,7 @@ For other content, you'll need to handle the [**DragStarting**](/uwp/api/windows
 ## Enable dropping
 
 The following markup shows how to set a specific area of the app as valid for dropping by using the [**AllowDrop**](/uwp/api/windows.ui.xaml.uielement.allowdrop) in XAML. If a user tries to drop somewhere else, the system won't let them. 
-If you want users to be able to drop items anywhere on your app, set the entire background as a drop target. Note that the background property defaults to null, which will not allow drops to happen (as there is nothing to hit-test), so make sure you set the background to a brush (a transparent brush will work too).
+If you want users to be able to drop items anywhere on your app, set the entire background as a drop target. Drops can only happen in areas where the `AllowDrop="True"` element receives pointer input. If the element doesn't have a background, you may need to set `Background="Transparent"` to ensure it receives input over its entire bounds.
 
 :::code language="xml" source="~/../snippets-windows/windows-uwp/design/input/drag_drop/cs/MainPage.xaml" id="SnippetDropArea":::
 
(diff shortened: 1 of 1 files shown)


hub/apps/design/input/drag-and-drop.md
```diff
@@ -51,7 +51,8 @@ For other content, you'll need to handle the [**DragStarting**](/uwp/api/windows
 
 ## Enable dropping
 
-The following markup shows how to set a specific area of the app as valid for dropping by using the [**AllowDrop**](/uwp/api/windows.ui.xaml.uielement.allowdrop) in XAML. If a user tries to drop somewhere else, the system won't let them. If you want users to be able to drop items anywhere on your app, set the entire background as a drop target.
+The following markup shows how to set a specific area of the app as valid for dropping by using the [**AllowDrop**](/uwp/api/windows.ui.xaml.uielement.allowdrop) in XAML. If a user tries to drop somewhere else, the system won't let them. 
+If you want users to be able to drop items anywhere on your app, set the entire background as a drop target. Note that the background property defaults to null, which will not allow drops to happen (as there is nothing to hit-test), so make sure you set the background to a brush (a transparent brush will work too).
```
codendone: The background _may_ be null, depending on whether the app has already set it for other reasons or whether the selected element is one which gets a background by default.

I suggest something like this:
```
Drops can only happen in areas where the `AllowDrop="True"` element receives pointer input. If the element doesn't have a background, you may need to set `Background="Transparent"` to ensure it receives input over its entire bounds.
```
If there is a doc about null/transparent to point to, even better. I didn't readily find anything.

asklar: thanks, good precision to add :)

asklar:
```suggestion
If you want users to be able to drop items anywhere on your app, set the entire background as a drop target. Drops can only happen in areas where the `AllowDrop="True"` element receives pointer input. If the element doesn't have a background, you may need to set `Background="Transparent"` to ensure it receives input over its entire bounds.
```

codendone (approved):

Karl-Bridge-Microsoft (approved):

asklar:
Hey @Karl-Bridge-Microsoft I feel like the final text doesn't really explain the fact that you _have_ to set the background to _something_ (i.e. don't leave it unset if you want to be able to drop onto it)

Karl-Bridge-Microsoft:
Hi @asklar - You said "may" in your original commit, which i removed to be more clear. Would you prefer I replaced it with "You must..."?

If the element doesn't have a background, you **may** need to set `Background="Transparent"` to ensure it receives input over its entire bounds.

asklar:
thanks @Karl-Bridge-Microsoft  yeah I think that wording makes it more explicit. In general, by default (*unless they inherit it from a parent/style/...) UI elements will have their background set to the empty brush (null), which means the user cannot drop onto them. If the developer wants users to drop onto the UI element, the developer must set the background to literally anything else, even a Transparent brush (which is different from null).

Karl-Bridge-Microsoft:
Thanks for the added detail @asklar! I've updated the topic and published. It should be live shortly.

Issue #5124, opened 2024-09-05
Link https://learn.microsoft.com/de-de/microsoft-store/working-with-line-of-business-apps is broken


regnete:
### Type of issue

Outdated article

### Feedback

The link https://learn.microsoft.com/de-de/microsoft-store/working-with-line-of-business-apps used in the docs is broken

### Page URL

https://learn.microsoft.com/de-de/windows/apps/publish/distribute-lob-apps-to-enterprises

### Content source URL

https://github.com/MicrosoftDocs/windows-dev-docs/blob/docs/hub/apps/publish/distribute-lob-apps-to-enterprises.md

### Author

@GrantMeStrength

### Document Id

b339e3ed-0cfe-12fe-c3a5-5a15082e126f

GrantMeStrength:
Hi Ankur  - Hoping you can take a look at this issue. Thanks!

PR #5146, opened 2024-09-28
Add a notice of retirement for lob-apps distribution

zaochih:
Add a notice as a temporary fix for #5124 before rewriting this document


GrantMeStrength:
Thank you for this update!

PR #5161, opened 2024-10-10
Fix typo in app-capability-declarations.md

cnihelton:
`packagesServices` seems a typo for `packagedServices`, which is the actual capability name.


drewbatgit (approved):

drewbatgit:
@CarlosNihelton Thanks for catching this!

PR #5167, opened 2024-10-13
Update uri-schemes.md

HO-COOH:


alvinashcraft (approved):
Nice catch, @HO-COOH. Thanks for your contribution.

PR #5205, opened 2024-11-14
Improved Advanced Paste docs - Custom Actions and Paste as .html

drawbyperpetual:


hub/powertoys/advanced-paste.md
```diff
@@ -31,7 +31,7 @@ From the Settings menu, the following options can be configured:
 | Clipboard history | Enable to automatically save clipboard history. |
 | Open Advanced Paste shortcut | The customizable keyboard command to open the **Advanced Paste** window. |
 | Automatically close the Advanced Paste window after it loses focus | Determines whether the Advanced Paste window will close after focus is lost from the window. |
-| Custom Actions | Create and manage advanced paste custom actions. (Requires Paste with AI to be enabled). |
+| Custom Actions | When using Paste with AI, save the prompts you frequently use and give them descriptive names so you can easily pick them out from the Advanced Paste window without having to type them out. You can also assign them keyboard commands so you can execute them without opening the Advanced Paste window. |
```
alvinashcraft:
```suggestion
| Custom Actions | When using Paste with AI, save the prompts you frequently use and give them descriptive names, so you can easily select them from the Advanced Paste window without having to type them out. You can also assign each action a keyboard command, so you can execute them without opening the Advanced Paste window. |
```

hub/powertoys/advanced-paste.md
```diff
@@ -118,24 +118,15 @@ If pasting files is accepted within the application that you are using (e.g. Fil
 
 ### Paste as .html file
 
-**Paste as .html file** enables you to paste text stored in your clipboard as a .html file with an auto-generated file name. You can optionally set a quick key shortcut in settings.
+**Paste as .html file** enables you to paste html data stored in your clipboard as a .html file with an auto-generated file name. This is especially useful for saving a part of a web-page from a browser - including links, formatted text and images. You can optionally set a quick key shortcut in settings.
```
alvinashcraft:
```suggestion
**Paste as .html file** enables you to paste html data stored in your clipboard as a .html file with an auto-generated file name. This is especially useful for saving a part of a webpage from a browser - including links, formatted text and images. You can optionally set a quick key shortcut in settings.
```

hub/powertoys/advanced-paste.md
```diff
@@ -118,24 +118,15 @@ If pasting files is accepted within the application that you are using (e.g. Fil
...
-<head>
-    <title>PowerToys</title>
-</head>
-<body>
-    <h1>PowerToys</h1>
-    <p>PowerToys is a set of utilities for power users to tune and streamline their Windows experience for greater productivity.</p>
-</body>
-</html>
-```
+<pre>
```
alvinashcraft: Why remove the markdown syntax block? We should only use HTML in the .md pages when there's no markdown equivalent.

drawbyperpetual: The markdown syntax block is inappropriate for Paste as .html. When you hit the "copy" button for the generated code-block, it copies *text* to the clipboard. But Paste as .html needs HTML data, not text (even text that corresponds to HTML code won't do the trick unfortunately). If you feel strongly against the `<pre>` block, then we should remove the sample here entirely.

alvinashcraft: That's a good point. I don't think it makes sense to have a Sample Input block for this section. Best to remove it. The info you added to the description gives the use an idea of how it can be used.

drawbyperpetual: Done

alvinashcraft (approved):
Looks great. Thanks again for updating this to improve the docs!

PR #5215, opened 2024-11-20
Update to the new WinUI logo

niels9001:


alvinashcraft:
@niels9001 There's an old-style logo in the banner in the WinUI section of the [Overview of framework options](https://learn.microsoft.com/windows/apps/get-started/#winui) page too. Should we take care of that separately?

niels9001:
> @niels9001 There's an old-style logo in the banner in the WinUI section of the [Overview of framework options](https://learn.microsoft.com/windows/apps/get-started/#winui) page too. Should we take care of that separately?

Great catch! Updated that one as well!

alvinashcraft (approved):
Looks good. I'll let you know if I spot any others.

PR #5240, opened 2024-12-12
applifecycle-single-instance: improve the MSBuild sample

ForNeVeR:
- No need to define the same property for each configuration/platform pair, should be possible to update globally
- Previous sample was causing losing of the previous content of DefineConstants, the new one is not


alvinashcraft (approved):
Thanks for this update. I tested it in my local project, and everything works as expected.

PR #5258, opened 2025-01-07
Update nodejs-beginners-tutorial.md to fix error

justinzw:
I ran through the tutorial. The issue with using  `echo var msg > app.js` is that by default it created a UTF-16 format file which then created extra characters and node did not like.

I changed the tutorial around so that you create the file "app.js" inside of VS code instead of `echo` to create it.

By default this will create the file in the right format.


mattwojo:
Love the solution. Thank you for this update @justinzw! Merged.

PR #5260, opened 2025-01-08
In the new store, it is used for apps as well

crramirez:


GrantMeStrength:
I suspect this entire section is out of date, as UWP apps are not supported for the Xbox any more.
Investigating.

PR #5269, opened 2025-01-15
Set BindingContext for Data Binding in MainPage Constructor

bhavanesh2001:
This ensures that the BindingContext of MainPage is properly set to the corresponding ViewModel instance.


bhavanesh2001:
#sign-off

alvinashcraft (approved):
This looks good. Thanks for catching this omission!

PR #5276, opened 2025-01-21
Update nodejs-on-wsl.md

jrn1993:
Settings Sync extension deprecation. Microsoft built-in **Settings Sync is On** as comparable solution.


mattwojo:
Thank you for catching this change and contributing the update, @jrn1993 !!

PR #5300, opened 2025-02-11
[DRAFT][Accessibility] Document the HighContrastChanged fallback

citelao:
Today, WinAppSDK devs should be using [`ThemeSettings.Changed`](https://learn.microsoft.com/en-us/windows/windows-app-sdk/api/winrt/microsoft.ui.system.themesettings.changed?view=windows-app-sdk-1.4) to detect high-contrast changes, not [`AccessibilitySettings.HighContrastChanged`](https://learn.microsoft.com/en-us/uwp/api/windows.ui.viewmanagement.accessibilitysettings.highcontrastchanged?view=winrt-26100) (which depends on CoreWindow).

This change documents that requirement.

See [corresponding docs change on `HighContrastChanged`](https://github.com/MicrosoftDocs/winrt-api/pull/2491) & [issue in WinAppSDK codebase](https://github.com/microsoft/microsoft-ui-xaml/issues/4163

> [!WARNING]
> HOLD until we validate this is a complete replacement.


stevewhims (approved):

scfeng10:
It seems that while this is the correct API to use to listen to theme changes, the Changed event does not fire when switching between High Contrast themes. I filed an internal bug to track this issue [Bug 56208821](https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems/edit/56208821): ThemeSettings.Changed event does not fire when switching between High Contrast themes

PR #5350, opened 2025-03-17
Add `sq` prefix tip for unit converter

PesBandi:
Docs PR for https://github.com/microsoft/PowerToys/pull/37862


alvinashcraft (approved):
Nice tip. Thanks for adding it. I'll hold onto this until the feature is released.

PR #5447, opened 2025-05-20
Update color-picker.md

skanda890:


alvinashcraft (approved):
Thanks for this fix! Looks good.

PR #5478, opened 2025-06-09
Update creating-an-extension.md

ackvf:
Make it instantly obvious what language is used to develop extensions.


alvinashcraft (approved):
That works. Thanks for making things clearer.

PR #5528, opened 2025-07-20
[Powertoys] Add screenshot mode to crop-and-lock.md

fm-sys:
Powertoys PR: https://github.com/microsoft/PowerToys/pull/40720

### DIFF
--- a/hub/powertoys/crop-and-lock.md
+++ b/hub/powertoys/crop-and-lock.md
@@ -41,6 +41,10 @@ Creates a window that replaces the original window, showing only the selected ar
 
 Not every window will react well to being contained in another application so this mode has many compatibility issues. It's advisable to use the "Thumbnail" mode instead if you find that a window isn't responding well to being cropped with the "Reparent" mode.
 
+### Screenshot
+
+Creates a window that shows the selected area of the original window as a freezed snapshot. This mode has good compatibility as well.
+
 ## Known issues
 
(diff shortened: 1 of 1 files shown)


hub/powertoys/crop-and-lock.md
```diff
@@ -41,6 +41,10 @@ Creates a window that replaces the original window, showing only the selected ar
 
 Not every window will react well to being contained in another application so this mode has many compatibility issues. It's advisable to use the "Thumbnail" mode instead if you find that a window isn't responding well to being cropped with the "Reparent" mode.
 
+### Screenshot
+
+Creates a window that shows the selected area of the original window as a freezed snapshot. This mode has good compatibility as well.
```
alvinashcraft: What does the second sentence mean? The mode is compatible with supported versions of Windows 10 and Windows 11? Or is it compatibility with different types of apps? I'd like to be more specific about what kind of compatibility we're talking about here. Thanks!

```suggestion
Creates a window that shows the selected area of the original window as a frozen snapshot. This mode has good compatibility as well.
```

fm-sys: I've now clarified that it's about compatibility among apps...

GrantMeStrength:
Wondering about the status of this PR?

fm-sys:
From my side it would be ready, don't know about the internal processes though :)

niels9001 (approved):

PR #5543, opened 2025-07-30
[x:Bind] Document TargetNullValue behavior with functions

citelao:
Today, we don't document how x:Bind behaves when you use functions & `TargetNullValue` together:

```
"{x:Bind local:ConverterFunctions.QualityTypeToColor(ViewModel.QualityType), Mode=OneWay, TargetNullValue='#FF808080'}"
```

This documents that the `TargetNullValue` is applied if the function returns a null, not if the arguments are null.

### DIFF
--- a/uwp/data-binding/function-bindings.md
+++ b/uwp/data-binding/function-bindings.md
@@ -170,6 +170,9 @@ Multiple function arguments can be specified, separated by comma's (,)
 - Constant Number - for example -123.456
 - Boolean – specified as "x:True" or "x:False"
 
+> [!TIP]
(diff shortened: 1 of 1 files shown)


alvinashcraft (approved):
Thanks for this addition to the doc. There's [another topic](https://learn.microsoft.com/en-us/windows/apps/develop/data-binding/function-bindings) that's nearly identical, but it frames the topic for WinUI apps instead of UWP. I suspect the same tip would be helpful there too.

I need to consolidate these two sets of data binding articles and just call out the differences between WinUI and UWP inline. Maintaining both areas isn't effective, and it's probably confusing for readers.

PR #5558, opened 2025-08-12
Update Windows App SDK downloads.md

michael-hawker:
Make it clearer at the bottom of the table where people will be looking that there's an archive page with prior releases.

I totally didn't see the link in the opening paragraph or the sidebar page listing initially, I had just gone to the table looking for the version I needed and didn't find it.

Having a link at the bottom of the table should make it easier to clue-in that there's more elsewhere.

This is important as old messages requiring older installs still dump users on the main page, and they may need to get the exact version they require in order to install the application they're trying to use.


Karl-Bridge-Microsoft (approved):

michael-hawker:
Thanks @Karl-Bridge-Microsoft! Appreciate the extra updates! Have a great day!

PR #5585, opened 2025-08-20
Replace deprecated Microsoft Store URLs with new format

zaochih:
Replace all Microsoft Store URLs with new format


Karl-Bridge-Microsoft:
@GrantMeStrength - can you review?

GrantMeStrength (approved):
Looks like links are getting mapped on the fly from the "old" format to the "new", so this chance makes sense.

PR #5589, opened 2025-08-22
Fix 1.8pre release note anchor

riverar:
Addresses microsoft/WindowsAppSDK#5727


Karl-Bridge-Microsoft:
Fixed in a private PR

PR #5641, opened 2025-09-08
docs: remove invalid `path =` line from MarkdownContent sample

BitRyuu:
### Summary
This PR fixes the MarkdownContent example by removing a dangling `path =` line.

### Details
- The current docs contain a `path =` line in the constructor that will not compile.
- `ContentPage` does not define a `path` property, so the line is invalid.
- The example already uses `doc_path` properly with `OpenUrlCommand`.

### Fix
- Removed the invalid line.
- Verified that the example now compiles cleanly.


alvinashcraft (approved):

alvinashcraft:
Thanks for the contribution. You're right that this looks like the line was mistakenly added.

PR #5642, opened 2025-09-08
Updated docs to indicate lack of support for school accounts

desigrit:
Changed a word from "Enterprise" to "School" to indicate that the portable web installer is NOT going to run for apps that are >E rating for users using school accounts.


desigrit:
@GrantMeStrength ready for your stamp of approval! :)

PR #5679, opened 2025-09-26
fixed inconsistent sizing for segoe-mdl docs images

ahrGNUts:
Was looking thru [this documentation](https://learn.microsoft.com/en-us/windows/apps/design/style/segoe-ui-symbol-font) today and saw this happening in a few places on the page: 
sizing

Scaled down the offending images so everything is 32x32px.


ahrGNUts:

GrantMeStrength:
Thanks for catching this! You're right that these 6 images are 64×64 while the rest are 32×32. We'll fix this internally — the file paths have been restructured in our working branch, so this PR can't be merged directly. Closing in favor of an internal fix. Thank you for the contribution!

PR #5680, opened 2025-09-28
Update WinUI 3 Gallery links

marcelwgn:
The WinUI 3 Gallery expects links to be of the form "winui3gallery://item..." with two "/". Most of the links only have one "/" resulting in the page not being opened. This PR fixes that so the links properly work.


jwmsft (approved):

jwmsft:
@marcelwgn, thanks for these fixes. Sorry I missed this PR until now. We've moved these files to a new location resulting in conflicts with this PR, so I've made these fixes in a new PR and will close this one.

PR #5695, opened 2025-10-12
Update web-frameworks.md python or python3 in cmd?

nericat2005:
before the section ##Hello World tutorial for Flask, I tried running "python filename.py" but it didn't work. Instead, "python3 filename.py" worked.


mattwojo:
Thanks for the doc update on this!

PR #5700, opened 2025-10-16
Remove MAT articles, update announcements

joshftb:
MAT is now deprecated, so we should remove the download links and product overviews.


Karl-Bridge-Microsoft:
Move to internal PR.

Karl-Bridge-Microsoft:
updated in a private PR

PR #5722, opened 2025-10-31
Add FAQ entry for x64 shell extension DLL on Arm64

cts-davean:
64-bit shell extension DLLs installed on Arm64 devices should be built for Arm64X to support loading in both Arm64 and x64 processes.

### DIFF
--- a/landing/arm-docs/faq.yml
+++ b/landing/arm-docs/faq.yml
@@ -65,6 +65,11 @@ sections:
 
           * `C:\Program Files (Arm)`: This directory was used for 32-bit Arm applications, which are [no longer being supported in future versions of Windows](./arm32-to-arm64.md).
 
+      - question: |
+          Will an x64 shell extension DLL work in Windows Explorer on Windows 11 Arm64 devices?
+        answer: |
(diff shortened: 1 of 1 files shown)


mattwojo:
Thanks for contributing this @cts-davean. I shared with the dev team and they said that we should add a bit of additional context about how shell extensions work (in-process vs out-of-process). I'm going to follow up to see if we can capture some of that nuance in a concise way to include in this FAQ before we merge it in.

cts-davean:
@mattwojo Perhaps the following will work.

[Shell extensions](https://learn.microsoft.com/windows/win32/shell/shell-exts) are in-process Component Object Model (COM) objects implemented as DLLs. On Windows 11 Arm64 devices, File Explorer (explorer.exe) runs as a native Arm64 process and cannot load x64 DLLs. While shell extensions may also be loaded by other processes that present a File Explorer-like view, such as applications displaying Open or Save As dialogs, shell extensions installed on Arm64 systems should be built for Arm64X. This ensures compatibility, allowing the DLL to load in both Arm64 and x64 processes. See [Build Arm64X binaries](./arm64x-build.md).

PR #5784, opened 2026-01-10
Troubleshooting note for App Installer

kkaminsk:
Not a big change but could help if the device isn't quite up to date.


GrantMeStrength:
Thanks! Looks useful!

PR #5837, opened 2026-02-25
Reflecting recent changes about installerTypes from https://github.com/microsoft/winget-cli/blob/master/doc/Settings.md

DandelionSprout:
Directly based on https://github.com/microsoft/winget-cli/pull/6067.

The current documentation on the priority orders of installer types when using `winget install` without having specified `--installer-type`, was unfortunately pretty scarce. So I've filled in most of the gaps now.

### DIFF
--- a/hub/package-manager/winget/settings.md
+++ b/hub/package-manager/winget/settings.md
@@ -69,6 +69,8 @@ The code snippet below shows an example of what your settings file might look li
 }
 ```
 
+In particular, the `"$schema": "https://aka.ms/winget-settings.schema.json",` line is required in order for any of the file's settings to be applied.
+
 We have also defined a schema for the settings file. This allows you to use TAB to discover settings and syntax if your JSON editor supports JSON schemas.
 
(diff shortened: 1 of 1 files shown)


hub/package-manager/winget/settings.md
```diff
@@ -69,6 +69,8 @@ The code snippet below shows an example of what your settings file might look li
 }
 ```
 
+In particular, the `"$schema": "https://aka.ms/winget-settings.schema.json",` line is required in order for any of the file's settings to be applied.
```
JohnMcPMS: This should not be true; I would consider it a bug if you have evidence of it.

DandelionSprout: It was at least what I concluded with yesterday (Tuesday) after various tests, but I'll see if I have time tomorrow to doublecheck it.

DandelionSprout: Huh. Turns out you're right after all. My fault.

PR #5847, opened 2026-03-03
Fix typo in floating point optimization description

cts-davean:
Fix typo in floating point optimization description


GrantMeStrength:
thanks!

PR #5895, opened 2026-04-09
Update Modern .NET UWP app docs

tommcdon:
Update recent UWP doc publish a Tip fix and removal of erroneous text stating that NativeAOT is required for store publish

### DIFF
--- a/uwp/dotnet-native/modernize-uwp-apps-with-dotnet.md
+++ b/uwp/dotnet-native/modernize-uwp-apps-with-dotnet.md
@@ -111,7 +111,7 @@ To migrate an existing UWP app from [.NET Native](index.md) to modern .NET:
 ### Step 1: Update Project File
 
 1. Convert your existing .csproj to SDK-style format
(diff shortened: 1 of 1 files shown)


uwp/dotnet-native/modernize-uwp-apps-with-dotnet.md
sylveon: Should probably change this too?

Something like
> Add the required properties (`UseUwp`, `EnableMsixTooling`). Optionally, add `PublishAot`.

ahmed605: I think something like
> 2. Add the required properties (`UseUwp`, `EnableMsixTooling`, and `PublishAot` or `SelfContained`)

would be better

tommcdon: I changed it to `UseUwp`, `EnableMsixTooling`, then `PublishAot` or `SelfContained`, to make it clear that we need either PublishAot or SelfContained

stevewhims (approved):

